PR #8, opened 2017-10-05
VAM-9 Load size

sidjtd:
Async ftw!


lib/broker/models/orwch-loadsiz.js
```diff
@@ -0,0 +1,44 @@
...
+        ],
+        invokeRPC: (args, session, callback) => {
+            if (_.isEmpty(session.userId)) {
+                callback(null, '');
+                return;
+            }
+            // Get the font size via ParameterService
+            testService.loadSize(args, session, callback);
+        },
+        formatResult: res => res,
```
mfuroyama: So, if the `formatResult` function does nothing, you can omit it...in the Pure JS implementations, these are optional

lib/services/testService/loadSize.js
```diff
@@ -0,0 +1,32 @@
...
+const ParameterService = require('../parameterService');
+
+/**
+ * Get the font size via ParameterService
+ * @param  {Object} context - the session data
+ * @param  {Function} callback - the callback function used to process results
+ */
+function loadSize(args, context, callback) {
+    const name = args[0] || false;
+    ParameterService.withContext(context).get('ORWCH BOUNDS', { entity: 'USR', instance: name})
```
mfuroyama: Woof...I hate to be pedantic about things like this, but don't forget the space after `name`...

lib/services/testService/loadSize.js
```diff
@@ -0,0 +1,32 @@
...
+        .then((res) => {
+            if (_.isString(name) && _.has(res, 'value')) {
+                let val = (_.isArray(res) && res.length > 0) ? res[0].value : res.value;
+                console.log('cuute', val);
```
mfuroyama: Also, even though this debug is 'chou kawaii' you might want to remove this from the final committed code set.  ;)

lib/broker/models/orwch-loadsiz.js
sidjtd: Got it!!!!

lib/services/testService/loadSize.js
sidjtd: Yesssirrr! <3

mfuroyama:
@sidjtd One more thing to remember is that if you include the Jira issue number in the pull request title, it'll link the pull request to the issue and it'll automatically move the issue to 'In Review'

mfuroyama:
@sidjtd Also, in the near future, let's move the Service code out of the 'testService' and into something like a `orderMetaService` instead. The code though, as it is, looks good.

sidjtd:
Ah, roger on the pull request, I'll actually re-do the other one right now!

PR #10, opened 2017-10-11
VAM-10 Load all

sidjtd:
Did all the mumpsy output in the formatResult area! Otherwise its just 3 arrays in 1 papa array

Also just saw your comment on moving these files into its own folder, will do from next implementation!


lib/services/testService/loadAll.js
```diff
@@ -19,6 +20,7 @@ function loadAll(context, callback) {
             return ParameterService.withContext(context).get('ORWCH COLUMNS', { entity: 'USR' });
         }).then((res) => {
             aggregate.push(res);
+            console.log(aggregate);
```
mfuroyama: Hmm...you might want to consider removing this for the final push

sidjtd: Oh yeah, I just added that in a second ago for the @return documentation and forgot to remove it! whoops!

mfuroyama:
Cool! Just like the other pull request, let's move the Service code out of the `testService` and into an `orderMetaService`.

We'll have to create the `orderMetaService` class (just pattern it after the `allergyMetaService` class) and an associated `orderMetaService` directory to put the API functions in...

sidjtd:
Got it! Will move it in...

lib/services/orderMetaService.js
```diff
@@ -0,0 +1,24 @@
...
+const BaseService = require('./baseService');
+
+class OrderMetaService extends BaseService {
+    constructor() {
+        super('OrderMetaService');
+    }
+    // This method does more more esoteric stuff, like:
+    // * Manually trigger an event on the service, using both methods
+    // * Context wrap
+    doSomethingMoreEsotericWithEvents(data, context, callback) {
```
mfuroyama: You can delete this function out...this was part of the original test

PR #12, opened 2017-10-12
Vam 7 gmv dll version

kekuso:
- just a quickie!


lib/broker/models/gmv-dll-version.js
```diff
@@ -1,50 +1,33 @@
...
 
 const RPC_EMULATOR_MODEL = [{
     id: 'GMV_DLL_VERSION',
     name: 'GMV DLL VERSION',
     function: 'DLL^GMVUTL8',
     parameters: ['GMV DLL VERSION'],
     files: [],
+    domain: 'CPRS',
+    labels: ['META', 'VITAL'],
+    services: [],
```
mfuroyama: Ahhh...sorry, could you add the appropriate labels here in the `services` field array? Thanks!

PR #14, opened 2017-10-13
VAM-51 MetaPureJS Orwdal32Def

sidjtd:
No orphaned console logs except the in the .catch!


lib/broker/models/orwdal32-def.js
```diff
@@ -0,0 +1,62 @@
...
+        function: 'DEF^ORWDAL32',
+        domain: 'ALLERGY',
+        description: 'Returns default values and list sets for Allergy ordering dialog.',
+        rpcType: 'READ',
+        complexity: 'DIFFICULT',
+        returnType: 'SINGLE VALUE',
+        files: ['120.83, 120.84'],
+        parameters: [],
+        labels: ['META'],
+        services: [],
```
mfuroyama: Ooh, forgot that I needed to be checking these things, but for the metaPureJS RPC emulation models, let's start populating the `services` field with the actual Service objects used in the RPC. In this case, it would be 'AllergyMeta'

lib/broker/models/orwdal32-def.js
```diff
@@ -0,0 +1,62 @@
...
+        inputParameters: [],
+        invokeRPC: (rpcArgs, session, callback) => {
+            if (_.isEmpty(session.userId)) {
+                callback(null, '');
+                return;
+            }
+            allergyMetaService.allergyDefList(rpcArgs, session, callback);
+        },
+        formatResult: (res) => {
+            if (_.has(res[0], 'top10Reactions')) {
```
mfuroyama: Consider bailing early here with a negative (`!_.has`) check. This way, you can move all the conditional code out of the indented block.

lib/broker/models/orwdal32-def.js
sidjtd: Thumbs up!!! Will do from now

lib/broker/models/orwdal32-def.js
```diff
@@ -0,0 +1,62 @@
...
+    {
+        id: 'ORWDAL32_DEF',
+        name: 'ORWDAL32 DEF',
+        function: 'DEF^ORWDAL32',
+        domain: 'ALLERGY',
+        description: 'Returns default values and list sets for Allergy ordering dialog.',
+        rpcType: 'READ',
+        complexity: 'DIFFICULT',
+        returnType: 'SINGLE VALUE',
+        files: ['120.83, 120.84'],
```
mfuroyama: Could you change this from a single string to an array of string values?  Thanks!

`['120.83, 120.84']` ==> `['120.83', '120.84']`

lib/services/allergyMetaService/allergyDefList.js
```diff
@@ -0,0 +1,44 @@
...
+
+const projectionObj = {
+    _id: 0,
+    top10Reactions: 1,
+};
+
+function allergyDefList(args, context, callback) {
+    /* Main implementation - query Mongo for specified properties and format the output */
+    this.find('GMRAllergySiteParameters', {
+        query: {
```
mfuroyama: 👍

lib/services/allergyMetaService/allergyDefList.js
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env node
+
+'use strict';
+
+/**
+ * ORWDAL32 DEF
+ * @description : Returns default values and list sets for Allergy ordering dialog.
+ * @returns {Array} : Array containining top10Reaction object'
+ */
```
mfuroyama: I know this might seem borderline pedantic, but it's good practice to drop your JSDoc comment right above the function it's associated with.

Also, don't forget that although the service function is implementing functionality derived from an RPC, it's not necessarily tied to the RPC here, so you can remove the `ORWDAL32 DEF` moniker

lib/broker/models/orwdal32-def.js
sidjtd: Yeah, this looks way better...

PR #15, opened 2017-10-13
Vam 50 orwdal32 clinuser

kekuso:
- hope I implemented the Promise chaining within a loop right...


lib/broker/models/orwdal32-clinuser.js
```diff
@@ -0,0 +1,30 @@
...
+const RPC_EMULATOR_MODEL = [
+    {
+        id: 'ORWDAL32_CLINUSER',
+        name: 'ORWDAL32 CLINUSER',
+        function: 'CLINUSER^ORWDAL33',
+        parameters: ['OR ALLERGY ENTERED IN ERROR'],
+        files: ['8930', '8930.3'],
+        domain: 'ALLERGY',
+        labels: ['META'],
+        services: [],
```
mfuroyama: Could you add the name of the service here (i.e. `allergyMeta`) for documentation purposes?

lib/services/allergyMetaService/isClinUser.js
```diff
@@ -0,0 +1,70 @@
...
+    /* Get the parameter for the current user entity */
+    ParameterService.withContext(context).get('OR ALLERGY ENTERED IN ERROR', { entity: 'USR' })
+        .then((paramUsrObj) => {
+            const val = parseInt(_.get(paramUsrObj[0], 'value'), 10);
+            if (val) return callback(null, val);
+            /* Determine whether current user is a member of any user classes that have parameters set.
+            Get all class memberships for current user, then iterate list, check parameters for those entities. */
+
+            /* Get IEN of current user, then get all the class memberships that user belongs to */
+            const userIEN = context.userId.split('-')[1];
```
mfuroyama: Let's consider creating a utility function here for accessing IENs from id values, then let's use that utility function consistently in the code. This way, when we transition over from `<file type>-<IEN>` ids to `urn:vista:443:<file type>:<IEN>` ids, we only need to change the utility function once, and we won't have to touch the service implementation code.

For now, the utility will split a `<file type>-<IEN>` value

kekuso: ah good idea, I'll create a utility! Want me to create a separate branch/pull-request or just tack it onto this one?

lib/services/allergyMetaService/isClinUser.js
```diff
@@ -0,0 +1,70 @@
...
+        .then((usrClassObj) => {
+            /* Go through all memberships, grab parameter for said class; if it exists, we logical
+            // OR all the results to produce a final result */
+            let res = '';
+            if (usrClassObj && usrClassObj.length) {
+                let userClassList = [];
+                if (!Array.isArray(usrClassObj[0].userClass)) userClassList.push(usrClassObj[0].userClass);
+                else userClassList = usrClassObj[0].userClass;
+                userClassList.forEach((userClass) => {
+                    ParameterService.withContext(context).get('OR ALLERGY ENTERED IN ERROR', { entity: userClass.id })
```
mfuroyama: This is partly my fault, since it's not documented well, but you can pass in an array of entities in the `entity` field of the `get`. So, something like might work:
```Javascript
const userClassIDs = userClassList.map(userClass => userClass.id);
ParameterService.withContext(context).get('OR ALLERGY ENTERED IN ERROR', { entity: userClassIDs })
...
```

lib/services/allergyMetaService/isClinUser.js
```diff
@@ -0,0 +1,70 @@
...
+                            callback(err);
+                        });
+                });
+            }
+
+            /* If we found something in user class check, return it */
+            if (res) return callback(null, res);
+
+            /* ...otherwise, get parameter for current division & system */
+            return ParameterService.withContext(context).get('OR ALLERGY ENTERED IN ERROR', { entity: context.division });
```
mfuroyama: You can probably take advantage of the entity array thing here too

lib/services/serviceUtils.js
```diff
@@ -0,0 +1,13 @@
...
+ * @summary Get an IEN from an ID value
+ * @param  {string} id - the id to split
+ * @return {string} IEN parsed from the id
+ * @example An ID value of '8989_51-50' would become '50'
+ */
+const getIenFromId = id => id.split('-')[1] || '';
+
+module.exports = {
+    getIenFromId,
+};
```
mfuroyama: 👍  🏅

PR #16, opened 2017-10-13
VAM-168: Update the MongoDB load utilities

mfuroyama:
This pull request is mostly for informational purposes. I need to move this branch in to get the Vagrant box generator working, but I wanted the team to be aware of the changes being made. These changes refine and fix a few issues related to the code that loads MVDM data into MongoDB.

The biggest issue addressed was the fact that the Mongoose initialization used by the MongoDB Data Access layer is asynchronous by nature. The impact of this is only felt when we're initializing a new database and if there are lots of schema models to load.  I fixed this by adding a "READY" event on the data access object, and wrapping the transfer functionality in a "READY" event handler.

This branch also:
* Adds additional schemas
* Cleans up a bunch of minor issues related to MVDM-to-schema mapping
* Tidies up the UX of the import script
* Cleans up some of the data access code


mfuroyama:
@kekuso It's like 85 schema files added and 4 or so actual code files that were updated

PR #22, opened 2017-10-19
VAM-188: Add RPC Model meta-field structure validity checker

creindle:
- [X] All required fields are present
- [X] All enumerated fields only contain allowable values
- [X] All field values are formatted properly, per the spec.
- [X] Use the tool to create a Jasmine test suite (with expect) that will fail on any QA errors
- [x] Need way to print which specific files failed


creindle:
Please write a lot! It's my first time using Jasmine specifically (I have used Mocha and Chai before), so I may not be utilizing all the best functionality!

lib/broker/models/tests/models-spec.js
```diff
@@ -0,0 +1,115 @@
...
+        }
+    });
+    it('rpcEmulatorModel should be defined', () => {
+        expect(rpcEmulatorModel).toBeDefined();
+    });
+});
+
+
+describe('rpcEmulatorModel', () => {
+    _.forEach(rpcEmulatorModel, (objInRPC) => {
```
mfuroyama: This is great! One thing I'd suggest is to have the spec execution output the name or ID of the RPC Model being tested. This way, if there's a failure, we'll know exactly which RPC was the offending one. There's several ways to do this:
1. Place the name in one or more of the string descriptions of the `it` function calls
2. Use nested `describes`, one for each RPC Model, with the RPC Model name as the description
3. `console.log`

mfuroyama: Okay, this is kinda pedantic, but a point about consistency in code: So technically, lodash `_.each` actually maps to `_.forEach` so they work, but for readability and consistency sake, I'd suggest picking either one and sticking with it throughout the spec module.

creindle: Yay! You're right! I added it in! I didn't realize it would be so easy!!

creindle: Thank you for the directions too!

creindle: Oooh... I could nest them... Hmm...

creindle: Oooh! I apologize! I though `_.forEach` was for arrays and `_.each` was for objects! I'll fix this!

PR #25, opened 2017-10-20
Use national ids

kekuso:
Did some cleanup and changed implementations to handle national IDs


lib/services/serviceUtils.js
```diff
@@ -60,7 +60,7 @@ const createIdObject = (id) => {
  * @return {Object} projObject - a projection list where all properties have a default value of 1 and _id: 0
  */
 const createProjection = (projections) => {
-    if (!Array.isArray(projections)) return { error: 'input must be an array' };
+    if (!Array.isArray(projections)) throw new Error('createProjection expected arg of type Array');
```
mfuroyama: Hmm...I'm a little on the fence about throwing an error here...let's discuss

PR #27, opened 2017-10-23
create test utils for service specs

kekuso:
for the segregation of model tests and service tests!


lib/services/tests/serviceTestUtils.js
```diff
@@ -0,0 +1,43 @@
...
+        onConnect: () => {
+            console.log('Connected in dataServiceSimple');
+        },
+    });
+
+    // initialize session values
+    const session = {
+        userId: '200-63',
+        division: '4-2957',
+        domain: '4_2-90',
```
mfuroyama: Cool! Thanks for doing this!

One minor thing: we might want to change these from the old-style Fileman IDs to the new-style URN national-type IDs

PR #30, opened 2017-10-25
VAM-97: ORWPCE GET EXAM TYPE - Implement the RPC Emulation Model Using Pure Javascript

creindle:


lib/broker/models/tests/unitTestUtils.js
```diff
@@ -8,7 +8,7 @@ function beforeAllScript(model) {
     DataAccess.getDefault().init({
         host: '10.2.2.222',
         port: 27017,
-        database: 'nodevista',
+        database: 'vam',
```
mfuroyama: Thank you for this!!

lib/broker/models/tests/orwpce-get-exam-type-spec.js
```diff
@@ -0,0 +1,26 @@
...
+        const res = unitTestUtils.beforeAllScript(rpcEmulatorModel);
+        rpc = res;
+    });
+
+    const testReturnsArray = (res) => {
+        expect(res).toBeDefined();
+        expect(res).toEqual(jasmine.any(Array));
+    };
+
+    it('Empty arg should return list', (done) => {
```
mfuroyama: Actually, here, you don't need to test the RPC `run`, just the `formatResult` function in the model; make sure the function works as expected. It's implied here that the `invokeRPC` function gets tested via the services tests

lib/services/pceMetaService/getExamType.js
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * @summary return an alphabetized list of active Exams
+ * @param  {string} args - deprecated. previously a date used for validation checking
+ * @param  {Function} callback - the function that will process results from getExamType
+ * @return {Array} res - array of Objects containing IEN and name properties (alphabetized by name)
+ */
+function getExamType(args, session, callback) {
+    // nodeVista local station number, to be changed depending on parameter
+    const localStationNumberParam = 6100;
```
mfuroyama: We want to make sure that we're using the `stationId` from the `session` context object. The idea here is that this object will hold all of a user's session information. For now, we assume the local station value will be set.

So, instead of hard-coding the variable here, let's change this to:
```Javascript
const localStationNumberParam = session.stationId;
```

lib/services/pceMetaService/getExamType.js
```diff
@@ -0,0 +1,36 @@
+'use strict';
+
+const serviceUtils = require('../serviceUtils');
+const _ = require('lodash');
+
+/**
+ * @summary return an alphabetized list of active Exams
+ * @param  {string} args - deprecated. previously a date used for validation checking
```
mfuroyama: Also, don't forget to document the `session` parameter.

lib/services/pceMetaService/getExamType.js
creindle: Ahh! Now I understand! I remember seeing the printout of "Session" and not connecting the dots to what it really was! Thank you for the specifics.

lib/broker/models/tests/unitTestUtils.js
creindle: It was all thanks to @kekuso  for figuring this out!

lib/broker/models/tests/orwpce-get-exam-type-spec.js
creindle: I understand what you're saying... I'm just a little bit stuck on how to test it as I know I would need the results from the service `invokeRPC` to pass into the function `formatResult()`... Hmmm. I'll ask after our morning scrum!

creindle: And I'm glad you addressed this as I was wondering where I was supposed to test this function!!

lib/services/pceMetaService/getExamType.js
```diff
@@ -6,22 +6,21 @@ const _ = require('lodash');
 /**
  * @summary return an alphabetized list of active Exams
  * @param  {string} args - deprecated. previously a date used for validation checking
+ * @param  {Object} session - user's session information: userId, division, domain, and stationId
  * @param  {Function} callback - the function that will process results from getExamType
- * @return {Array} res - array of Objects containing IEN and name properties (alphabetized by name)
+ * @return {Array} res - array of Objects containing IEN and label properties (alphabetized by label)
```
mfuroyama: Ahh, sorry...one more thing...could you add these two lines to the bottom of your service function JSDoc comments?
```
@memberof PCEMetaService
@instance
```

PR #35, opened 2017-10-26
VAM-99: ORWPCE GET HEALTH FACTORS TY

creindle:
- [X] All files created
- [X] Service spec is finished
- [X] Tests for both addCat and withoutCat
- [X] Passes model tests
- [X] Passes comparison tests


lib/broker/models/orwpce-get-health-factors-ty.js
```diff
@@ -0,0 +1,55 @@
...
+        }],
+
+        processRPCInvocation(rpcArgs, session, callback) {
+            pceMetaService.getHealthFactorsTY(rpcArgs, session, callback);
+        },
+
+        formatResult(res, cat) {
+            let result = res.map(element => `${element.ien}^${element.factor}`);
+
+            if (_.isUndefined(res) || !_.isArray(res)) {
```
mfuroyama: Let's put this guard up at the top of the function to make it the first thing that gets called.

This way, if `res` actually is undefined or if it's not an array, the call to `map()` above won't 💣 💥

lib/broker/models/orwpce-get-health-factors-ty.js
```diff
@@ -0,0 +1,55 @@
...
+        },
+        inputParameters: [{
+            id: 'ADDCATS',
+            parameterType: 'NUMERIC',
+            required: false,
+            description: 'Set this to 1 to retrieve both health factors *and* categories',
+            example: '1',
+        }],
+
+        processRPCInvocation(rpcArgs, session, callback) {
```
mfuroyama: Remember that the convention for the **metaPureJS** RPC emulation model function is `invokeRPC`. `processRPCInvocation` was the name we used in the old VDP models. Let's swap the name out here!

mfuroyama: Also, a good idea would be to run the most excellent model validity checker (via `npm run test-model-validity`) on your code...

lib/broker/models/orwpce-get-health-factors-ty.js
```diff
@@ -0,0 +1,55 @@
...
+            pceMetaService.getHealthFactorsTY(rpcArgs, session, callback);
+        },
+
+        formatResult(res, cat) {
+            let result = res.map(element => `${element.ien}^${element.factor}`);
+
+            if (_.isUndefined(res) || !_.isArray(res)) {
+                return new Error('res is either undefined or not array');
+            }
+            if (cat) {
```
mfuroyama: Ohhhh! I finally see what you're saying here! You want to pass formatting options.

So, the function signature requirement of `formatResult` is pretty strict: The upstream user of this function will only pass in one value, and that's the value that you returned in the second argument in the `invokeRPC` callback.

Consider a different alternative: maybe have `invokeRPC` return an object instead of an array. The object could contain the results and the formatting options. Let's discuss if you have questions about how to implement that.

lib/services/pceMetaService/getHealthFactorsTy.js
```diff
@@ -0,0 +1,36 @@
...
+function getHealthFactorsTY(args, session, callback) {
+    /* Get all active Exam IENs and labels */
+    this.find('HealthFactors', {
+        query: { inactiveFlag: { $exists: false }, localStationNumber: { $eq: session.stationId } },
+        projection: serviceUtils.createProjection(['factor', 'entryType', 'id', 'category']),
+    })
+        .then((examArr) => {
+            const res = _(examArr)
+                .sortBy(element => element.factor)
+                .map(el => ({ ien: serviceUtils.getIenFromId(el.id), factor: el.factor, entryType: el.entryType[0], categoryId: serviceUtils.getIenFromId(_.get(el, 'category.id', '')) }))
```
mfuroyama: This is really minor formatting issue, but normally, when defining objects using the nested object literal format, if there are more than, say, 3 or so fields, it's a good idea to break the object definition down on to multiple lines. So, for example:
```Javascript
.map(el => ({ ien: serviceUtils.getIenFromId(el.id), factor: el.factor, entryType: el.entryType[0], categoryId: serviceUtils.getIenFromId(_.get(el, 'category.id', '')) }))
```
becomes
```Javascript
.map(el => ({
    ien: serviceUtils.getIenFromId(el.id),
    factor: el.factor,
    entryType: el.entryType[0],
    categoryId: serviceUtils.getIenFromId(_.get(el, 'category.id', ''))
}))
```
It's subtle, but the second case makes things a lot more readable.

lib/broker/models/orwpce-get-health-factors-ty.js
```diff
@@ -0,0 +1,55 @@
...
+        function: 'HFTYPE^ORWPCE2',
+        domain: 'PCE',
+        description: 'Returns a list of active health factor types.',
+        rpcType: 'READ',
+        complexity: 'LOW',
+        returnType: 'ARRAY',
+        files: ['9999999.64'],
+        parameters: [],
+        labels: ['META'],
+        services: [],
```
mfuroyama: Don't forget to add the name of the service class you're using (in this case, `'pceMetaService'`) as a string here in the `services` array.

lib/services/pceMetaService/getHealthFactorsTy.js
```diff
@@ -0,0 +1,36 @@
...
+const _ = require('lodash');
+
+/**
+ * @summary return an alphabetized list of active Health Factors
+ * @param  {string} args - deprecated. previously a date used for validation checking
+ * @param  {Object} session - user's session information: userId, division, domain, and stationId
+ * @param  {Function} callback - the function that will process results from getExamType
+ * @return {Array} res - array of Objects containing IEN and label properties (alphabetized by label)
+ *
+ * @memberof ORWPCE GET HEALTH FACTORS TY
```
mfuroyama: Okay, here's where we get pedantic. Could you:
* Change this to `@memberof PCEMetaService`
* Add the `@instance` tag below

The `@memberof` tag ties this function with the 'PCEMetaService` class, defined in `services/PCEMetaService.js`, while the `@instance` tag tells the doc generator to not label this function as a static function. Thanks!

lib/broker/models/orwpce-get-health-factors-ty.js
creindle: I got it! Thank you!

lib/services/pceMetaService/getHealthFactorsTy.js
creindle: Awesome! It is much more readable, thank you for the tidbit.

lib/broker/models/orwpce-get-health-factors-ty.js
creindle: Oh... I didn't realize I needed to add this! I should probably check my other RPC?

PR #36, opened 2017-10-27
Vam 101 orwpce get set of codes

kekuso:
- create fileManService class and getFieldCodesFromFile method
- add some JSON files in services/fileManService/vdmNonClinicalModel
- create getSetOfCodes service and tests


lib/broker/models/orwpce-get-set-of-codes.js
```diff
@@ -0,0 +1,62 @@
...
+    {
+        id: 'ORWPCE_GET_SET_OF_CODES',
+        name: 'ORWPCE GET SET OF CODES',
+        function: 'GETSET^ORWPCE2',
+        domain: 'PCE',
+        parameters: [],
+        files: [],
+        labels: ['META', 'MULTIPLE-FILES'],
+        mvdm: {},
+        rpcType: 'READ',
```
mfuroyama: Hey, could you add a `services` array here, and add `'pceMetaService'` as an entry? Thanks!

lib/services/fileManService/getFieldFromFile.js
```diff
@@ -0,0 +1,46 @@
+'use strict';
+
+const vdmModels = require('../fileManService/vdmNonClinicalModel').vdmModel;
+
```
mfuroyama: You may want to consider pre-creating a map object here that has the VDM model file types (`fmId`) as keys. The lodash `transform` function is perfect for this:
```Javascript
const VDM_MODEL_MAP = _.transform(vdmModels, (map, model) => {
    map[model.fmId] = model;
}, {});
```
This way, you don't have to search through the entire `vdmModels` array on each call to `getFieldCodesFromFile`...just use the `VDM_MODEL_MAP`

lib/services/fileManService/getFieldFromFile.js
```diff
@@ -0,0 +1,46 @@
...
+ * @return {Object} setOfCodes Set of field values, or an error if not found
+ * @example
+ * getFieldCodesFromFile('200', '4') returns { 'NO TAKE': 'O', POSITIVE: 'P', NEGATIVE: 'N', DOUBTFUL: 'D' }
+ *
+ * @memberof fileManService
+ * @instance
+ */
+function getFieldCodesFromFile(fileId, fieldId) {
+    let setOfCodes = {};
+    vdmModels.some((model) => {
```
mfuroyama: See my comment above regarding the VDM model map. If you go that route, you can eliminate this top-level `some` call and clean up some of the logic.

lib/services/pceMetaService/getSetOfCodes.js
```diff
@@ -0,0 +1,27 @@
...
+const fileManService = require('../fileManService');
+
+/**
+ * @summary Returns values for a set of codes given a file and field number.
+ * @param  {Array} args First value is the fileID, second value is the fieldID, third (optional) value specifies whether or not an option should NOT be selected (1 - none selected, 0 - selected)
+ * @param  {Object} context Session context information associated with the service call.
+ * @param  {Function} callback Function called on process completion, with standard (err, result) signature.
+ * @return {Object} res An object containing a set of codes and a 'noneSelected' flag
+ *
+ * @memberof PceMetaService
```
mfuroyama: Okay, pedantic comment time, but I think needs to be `PCEMetaService`. Basically, it needs to match the value named in the `@class` tag of the `services/pceMetaServices.js` class file JSDoc comments.

lib/services/pceMetaService/getSetOfCodes.js
```diff
@@ -0,0 +1,27 @@
...
+ * @instance
+ */
+function getSetOfCodes(args, context, callback) {
```
mfuroyama: This isn't a big deal, but you can probably remove the `context` parameter here.

lib/services/fileManService/getFieldFromFile.js
kekuso: cool!

PR #41, opened 2017-10-28
Add serviceId to session/context values

sidjtd:
This PR is mostly for visibility, otherwise it's completely straight forward. More faux session data!


sidjtd:
Added to the wiki

PR #43, opened 2017-10-28
VAM-174 Model spec generating

sidjtd:
Maybe there is a more elegant solution to this, but now the generator creates a model-spec file if you include any 3rd argument string.  So now you can do:

 eg: >> npm run make -- p123_sumRpc
 eg: >> npm run make -- p123_sumRpc {customName}
 eg: >> npm run make -- p123_sumRpc {customName} {thisStringMakesMeAModelSpec}
 eg: >> npm run make -- p123_sumRpc {coolName} {abcd} 
eg: >> npm run make -- p123_sumRpc {coolName} {-} 

Last 3 lines out of the above will function in the exact same way.


scripts/metaBoiler.js
```diff
@@ -9,6 +9,7 @@ const boiler = require('./boilerTemplates');
 const userInput = process.argv[2] || '_';
 let myName = process.argv[3] || '';
 userInput.trim();
+const makeModelSpec = process.argv[3] || false;
```
mfuroyama: Hmm, did you mean this to be `process.argv[4]`?

mfuroyama: Also (and this is purely optional), you *could* implement n option checker, which would process all the command-line arguments after the required ones:
```Javascript
let makeModelSpec = false;

const options = process.argv.slice(4);
options.forEach((arg, index) => {
    if (arg === '-m' or arg === '--modelSpec') {
        makeModelSpec = true;
    }
});
````

sidjtd: I really like this, so I'm gonna take this if that's ok!

sidjtd: yup! Whoopsy

sidjtd: Actually that's not even in the current code anymore, really threw me off, that!

PR #44, opened 2017-10-30
Made session values uniform with entity map

creindle:
- [x] Double checking value for mock packageId


kekuso:
Hey Christie, can you do me a favor and update isClinUser in services/orderMetaService and include that change in this pull request?
I think it will break since I used 'domain' and 'division' instead of 'domainId' and 'divisionId',
You might want to check if there are any other parameterService implementations that break too :)

lib/services/allergyMetaService/getAllergyDefList.js
```diff
@@ -13,20 +13,22 @@ const serviceUtils = require('../serviceUtils');
...
     this.find('GMRAllergySiteParameters', {
         query: {
             $or: [{
                 name: 'HOSPITAL',
             }, {
-                division: { $elemMatch: { id: context.division } },
+                division: { $elemMatch: { id: context.divisionId } },
             }, {
-                id: `urn:vista:${context.stationId}:120_84:1`,
+                id: `urn:vista:${context.locationId}:120_84:1`,
```
mfuroyama: I'm thinking maybe we should leave this as `stationId`?

lib/services/tests/serviceTestUtils.js
```diff
@@ -16,10 +16,11 @@ function beforeAllScript(model) {
     // initialize session values
     const session = {
         userId: 'urn:vista:6100:200:63',
-        division: 'urn:vista:6100:4:2957',
-        domain: 'urn:vista:6100:4_2:90',
-        service: 'urn:vista:6100:49:1',
-        stationId: '6100',
```
mfuroyama: Let's leave `stationId` as-is, since it aligns with the terminology we're using upstream

PR #46, opened 2017-10-31
Made session object into JSON

creindle:
- [x] Need to pass all tests with `npm run test-comparison` (Getting errors with ORWCH LOADALL)


lib/broker/dispatchers/passthru.js
```diff
@@ -15,7 +15,7 @@ class PassthruDispatcher {
     constructor(config) {
         this.config = config || {};
         this.rpcRunner = new RPCRunner(GTM.db);
-        this.userId = this.config.userId || 62;
+        this.userId = this.config.userId || 63;
```
mfuroyama: We might have to change this back to 62, let's all discuss this tomorrow as a team

mfuroyama:
Yeah, I think the errors we're seeing are actually discrepancies between the MongoDB VM and the nodeVISTA v1.3.1 image. I'll regenerate the VM tonight and test with your branch.

PR #48, opened 2017-10-31
Refactor model spec tests

sidjtd:
Corrected model test specs for loadAll and getVitalType

Also removed all un-used parameters from test specs files, service API files, and in the parameters of the invocation of the serviceAPI within the model files.

JSDoc should also have matching information


lib/services/pceMetaService/getSetOfCodes.js
```diff
@@ -4,18 +4,17 @@ const fileManService = require('../fileManService');
...
  * @return {Object} res An object containing a set of codes and a 'noneSelected' flag
  *
  * @memberof PCEMetaService
  * @instance
  */
-function getSetOfCodes(args, callback) {
-    const orwFile = args[0] || '';
-    const orwField = args[1] || '';
-    const orwNull = args[2] || 0;
+function getSetOfCodes(fileFieldNull, callback) {
```
mfuroyama: Don't have to address this now (we'll come back to it), but this is a great example of how we can break Array type args out into individual parameters. This might be better served with the following signature:
```Javascript
function getSetOfCodes(fileType, fieldId, isNull, callback) {
```

lib/services/vitalsMetaService/getVitalTypeIEN.js
```diff
@@ -5,29 +5,26 @@ const serviceUtils = require('../serviceUtils');
...
+    this.find('GmrvVitalType', {
+        query: { label: vitalType[0] },
+        projection: serviceUtils.createProjection(['id']),
+    }).then((vitalObj) => {
+        if (_.has(vitalObj[0], 'id')) callback(null, vitalObj);
+        return callback(null, 0);
+    }).catch((err) => {
+        console.log(err);
+        return callback(err);
+    });
```
mfuroyama: 👍  Nice cleanup!

lib/services/pceMetaService/getSetOfCodes.js
sidjtd: Yes, much cleaner. Will do!

PR #49, opened 2017-10-31
Vam-85: ORWORB GETSORT

creindle:
- [X] Created initial files
- [x] Ensure ParameterService has all parameters 
- [x] Pass test-comparison


lib/services/orderMetaService/getSort.js
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * @memberof OrderMetaService
+ * @instance
+ */
+function getSort(context, callback) {
+    let aggregate = '';
+    ParameterService.withContext(context).get('ORB SORT METHOD', { entity: 'USR' })
+        .then((res) => {
+            aggregate += res[0].value;
+            aggregate += '^';
```
kekuso: this is optional, but could save one line!
`aggregate += `${res[0].value}^`;`

kekuso: You have no idea how long it took me to format that code in Markdown -_-

lib/broker/models/orworb-get-sort.js
```diff
@@ -0,0 +1,36 @@
...
+
+    {
+        id: 'ORWORB_GETSORT',
+        name: 'ORWORB GETSORT',
+        function: 'GETSORT^ORWORB',
+        parameters: ['ORB SORT METHOD', 'ORB SORT DIRECTION'],
+        files: [],
+        domain: 'ORDER',
+        labels: ['META'],
+        services: [],
```
kekuso: as Mike would say..."I hate to be pedantic but please include 'orderMetaService' here" :)

lib/broker/models/orworb-get-sort.js
```diff
@@ -0,0 +1,36 @@
...
+        mvdm: {},
+        rpcType: 'READ',
+        complexity: 'LOW',
+        description: 'Returns method for sorting GUI alert display.',
+        returnType: 'SINGLE VALUE',
```
kekuso: I think you can change 'SINGLE VALUE' to 'STRING' or 'LITERAL' depending on the output

lib/services/orderMetaService/getSort.js
```diff
@@ -0,0 +1,35 @@
...
+            return ParameterService.withContext(context).get('ORB SORT DIRECTION', { entity: 'USR' });
+        })
+        .then((res) => {
+            aggregate += res[0].value;
+            console.log(aggregate);
```
kekuso: can remove the console.log once everything is finalized

lib/broker/models/orworb-get-sort.js
mfuroyama: LOL! Shoot, I would say that...  🤔   Maybe I need to change things up a little...

lib/services/orderMetaService/getSort.js
mfuroyama: Actually, we'll talk about this later @creindle, but let's think about moving the output data formatting code out to the RPC model `formatResult` function.

So, you can think of each RPC implementation issue as having two distinct parts:
1. Get the data (in a generic form)
2. Format the data (in an RPC-specific form)

The service functions should be responsible for Part 1, while the RPC model `formatResult` function should be made responsible for Part 2.

Generally speaking, carets in the original RPC emulation outputs mean that the output is RPC-specific, so that formatting should be done in `formatResults`.

If this is confusing or if you have questions, we can discuss...

lib/broker/models/orworb-get-sort.js
```diff
@@ -0,0 +1,36 @@
...
+        inputParameters: [],
+        invokeRPC(rpcArgs, session, callback) {
+            orderMetaService.getSort(session, callback);
+        },
+
+        formatResult(res) {
+            return res;
```
mfuroyama: We should be doing the conversion of the results to RPC-specific format (i.e. `D^F`) here. We'll discuss...

lib/services/orderMetaService/getSort.js
creindle: Thanks for the catch!!

lib/broker/models/orworb-get-sort.js
```diff
@@ -27,7 +27,14 @@ const RPC_EMULATOR_MODEL = [
...
         formatResult(res) {
-            return res;
+            let aggregate = '';
+            if (res.length === 0) {
+                return '';
+            }
+            aggregate += res[0];
+            aggregate += '^';
+            aggregate += res[1] || '';
+            return aggregate;
```
mfuroyama: Here's a nifty little trick: instead of doing string concatenation, you could replace this entire block with:
```Javascript
return res.join('^');
```

For this to work, the service function would need to always return an array with two elements...

lib/services/orderMetaService/getSort.js
```diff
@@ -12,21 +12,21 @@ const ParameterService = require('../parameterService');
...
 function getSort(context, callback) {
-    let aggregate = '';
-    ParameterService.withContext(context).get('ORB SORT METHOD', { entity: 'USR' })
+    const aggregate = [];
+    ParameterService.withContext(context).get('ORB SORT METHOD', { entity: 'ALL' })
         .then((res) => {
-            aggregate += res[0].value;
-            aggregate += '^';
-            return ParameterService.withContext(context).get('ORB SORT DIRECTION', { entity: 'USR' });
+            console.log('res', res);
```
mfuroyama: These `console.log` statements should probably be removed

lib/broker/models/orworb-get-sort.js
```diff
@@ -0,0 +1,42 @@
...
+        invokeRPC(rpcArgs, session, callback) {
+            orderMetaService.getSort(session, callback);
+        },
+
+        formatResult(res) {
+            let aggregate = '';
+            if (res.length === 0) {
+                return '';
+            }
+            aggregate += `${res[0].value}^`;
```
mfuroyama: I get the sense that this might not work like how you expect it to, since, in the service function, you're already passing `res[0].value`. You might want to double check this...

lib/broker/models/orworb-get-sort.js
```diff
@@ -0,0 +1,42 @@
...
+            aggregate += res[1] || '';
+            return aggregate;
```
mfuroyama: If you're returning your results in an array, then consider replacing the string concatenation with a simple `join` call:
```Javascript
return res.join('^');
```

lib/services/orderMetaService/getSort.js
```diff
@@ -0,0 +1,40 @@
...
+ */
+function getSort(context, callback) {
+    const aggregate = [];
+    ParameterService.withContext(context).get('ORB SORT METHOD', { entity: 'ALL' })
+        .then((res) => {
+            if (res.length === 0) {
+                aggregate.push('');
+            } else {
+                aggregate.push(res[0].value);
+            }
```
mfuroyama: Consider DRY-ing the code a little by replacing the `if-else` here (and below) with a slightly more efficient pattern. There are several ways to do this, depending on your preference for readability. One pattern to consider is a ternary:
```Javascript
const value = (res.length > 0) ? res[0].value || '';
aggregate.push(value);
```

Another pattern to consider leverages lodash's fault tolerance to get the value out of the results:
```Javascript
const value = _.get(res[0], 'value', '');
aggregate.push(value);
```

lib/broker/models/orworb-get-sort.js
creindle: Yes... this was a mistake :O

lib/broker/models/orworb-get-sort.js
creindle: Yes! This was much better!

lib/services/orderMetaService/getSort.js
creindle: Awesome! The second pattern is really convenient!

lib/broker/models/orworb-get-sort.js
```diff
@@ -0,0 +1,39 @@
...
+        inputParameters: [],
+        invokeRPC(rpcArgs, session, callback) {
+            orderMetaService.getSort(session, callback);
+        },
+
+        formatResult(res) {
+            if (res.length === 0) {
+                return '';
+            }
+            return res.join('^');
```
mfuroyama: This is fine as is, but you might want to consider refactoring this down even further using a ternary:
```Javascript
return (res.length > 0) ? res.join('^') : '';
```
Up to you though...just some food for thought.

creindle: @.@ This is a very good option, I'm a little ashamed I missed the opportunity to apply a ternary here too!

lib/services/tests/getSort-spec.js
```diff
@@ -0,0 +1,29 @@
...
+     * ORWORB GET SORT
+     * @description return the notification
+     */
+
+    const results = {};
+
+    it('Empty arg should return integer', (done) => {
+        modelFunction.getSort(context, (err, res) => {
+            if (err) console.log(err);
+            results.empty = _.isArray(res);
```
mfuroyama: I just realized now: this test spec doesn't really test anything. 😄 

If an error pops up, you can manually fail the spec by calling the Jasmine `fail('msg')` function. Also, let's throw at least one `expect` at the results array...

creindle: Let me add more tests!

creindle: I created this file when I first received this VAM, so it slipped my mind to add more tests once I understood what the RPC needed in its response.

PR #52, opened 2017-11-01
VAM-222: Improve Parameter Service

mfuroyama:
As we discussed, this thread adds the following improvements to the Parameter Service:

* Parameter Definitions are now retrieved and used to determine the proper sort-order of the results. This will show through when requesting parameters with multiple entity options
* Parameter Definitions are cached to prevent the need for consecutive accesses to the database
* Cleaned up the parameter query builder and the entity mappings
* The 'ALL' entity type is now supported
* Added unit tests for the service
* Added improved JSDoc comments

These changes should be transparent to the users of the Parameter Service, but should make the output (and the documentation) much more consistent.


mfuroyama:
Gonna merge this in to give Christie access to the updated functionality

PR #53, opened 2017-11-02
Vam 182 orwu valdt

kekuso:
added validateDate and xsdToFmDate service methods and tests


lib/broker/models/dateTimeUtilitiesModel.js
```diff
@@ -5,6 +5,40 @@
 const dateTimeService = require('../../services/dateTimeService');
 
 const RPC_EMULATOR_MODEL = [{
+    id: 'ORWU_VALDT',
```
mfuroyama: Ahh, okay, here goes Mr. Pedantic again, but I know Conor's gonna balk at this:

I know this is tedious, but could we split this model up into individual model JS files? Conor likes the separation for visual tracking... 🤢  Thanks!

kekuso:
oh wait, should I split up the test specs too?

mfuroyama:
Nah, just the models. I/we like the models separated out one model-per-file for quick visual tracking. As long as the tests test the individual model formatting, the specs are fine in one file.

PR #58, opened 2017-11-03
VAM-98: ORWPCE GET EXCLUDED

creindle:
- [X] Created all files
- [X] Passed comparison tests
- [X] Correct JSDocs


lib/broker/models/orwpce-get-excluded.js
```diff
@@ -40,11 +40,14 @@ const RPC_EMULATOR_MODEL = [
...
-        processRPCInvocation(rpcArgs, session, callback) {
+        invokeRPC: (rpcArgs, session, callback) => {
             pceMetaService.getExcluded(rpcArgs, session, callback);
         },
 
-        toReturnValue(res) {
+        formatResult: (res) => {
+            if (_.isUndefined(res) || !_.isArray(res)) {
+                return new Error('res is undefined or not array');
+            } return res ? _.get(res, 'list', []) : [];
```
mfuroyama: Minor pedantic styling issue: could you drop this `return` to the next line? It makes the code slightly more readable...

lib/services/pceMetaService/getExcluded.js
```diff
@@ -6,36 +6,48 @@ const _ = require('lodash');
...
-    console.log('context', context);
-    let serviceSectionId;
-    // VDM.userId() - NewPerson get the id based on the context.userId
-    this.find('NewPerson', {
-        query: { id: { $eq: context.userId } },
-    })
-        .then((test) => {
-            // Get the service section
-            // console.log('service section id', ));
+function getExcluded(rpcArgs, context, callback) {
```
mfuroyama: One thing to remember is that the service functions shouldn't have any trace of the term 'RPC' in them.

Also, in general, we should be as explicit and descriptive as possible when it comes to service function parameters. Here, let's replace the single `rpcArgs` array parameter with something like `location, type`

lib/services/pceMetaService/getExcluded.js
```diff
@@ -6,36 +6,48 @@ const _ = require('lodash');
...
+                    'IMMUNIZATIONS',
+                    'SKIN TESTS',
+                    'PATIENT ED',
+                    'HEALTH FACTORS',
+                    'EXAMS',
+                ];
+    const param = loc ? typeNames[type - 1] : '';
+    if (!param) {
+        callback(null, []);
+        return [];
```
mfuroyama: No need to return anything, just a straight up `return` is fine... :)

lib/services/pceMetaService/getExcluded.js
```diff
@@ -6,36 +6,48 @@ const _ = require('lodash');
...
+    if (loc !== context.locationId[context.locationId.length - 1]) {
+        usrInputtedLoc = context.locationId.slice(0, context.locationId.length - 1) + loc;
+    }
+    const entArr = [
+        context.userId,
+        context.stationId,
+        context.serviceId,
+        context.divisionId,
+        context.domainId,
+        usrInputtedLoc];
```
mfuroyama: More minor formatting issues: let's drop the trailing bracket down to the next line. (weird why ESLint didn't catch this...)

lib/services/pceMetaService/getExcluded.js
```diff
@@ -6,36 +6,48 @@ const _ = require('lodash');
...
+    const typeNames =
+                [
+                    'IMMUNIZATIONS',
+                    'SKIN TESTS',
+                    'PATIENT ED',
+                    'HEALTH FACTORS',
+                    'EXAMS',
+                ];
```
mfuroyama: Also, another formatting issue: let's make the way we do our array assignments and bracket placements consistent here. Consider moving the starting bracket on the same line as the `const typeNames =` line and re-indenting accordingly

lib/services/pceMetaService/getExcluded.js
```diff
@@ -6,36 +6,48 @@ const _ = require('lodash');
...
-        .then((test) => {
-            // Get the service section
-            // console.log('service section id', ));
+function getExcluded(rpcArgs, context, callback) {
+    const loc = _.toNumber(rpcArgs[0]) || '';
+    const type = _.toNumber(rpcArgs[1]) || '';
+    let usrInputtedLoc;
+    if (loc !== context.locationId[context.locationId.length - 1]) {
+        usrInputtedLoc = context.locationId.slice(0, context.locationId.length - 1) + loc;
+    }
```
mfuroyama: Notice here: depending on the value of `loc` there's a chance that you'll be passing an `undefined` value in to `entArr` below. Consider either adding a guard or modifying the assignment logic to prevent this case.

lib/services/pceMetaService/getExcluded.js
```diff
@@ -6,36 +6,48 @@ const _ = require('lodash');
 
 /**
  * @summary returns a list of excluded PCE entries
+ * @param  {Array} rpcArgs - [location (ien) and select specific item to search for]
```
mfuroyama: ...and when you change the parameters in the function signature below, don't forget to change the JSDoc comments too! Thanks!

lib/services/tests/getExcluded-spec.js
```diff
@@ -18,15 +17,38 @@ describe('ORWPCE GET EXCLUDED asynchronous specs', () => {
      * parameters for the user in an array
      */
 
-    // const results = {};
+    it('Empty arg should return empty array', (done) => {
+        modelFunction.getExcluded([], context, (err, res) => {
+            if (err) console.log(err);
```
mfuroyama: I admit, I've been glazing over the test specs, but I'll need to be more thorough:

Let's manually `fail()` here if we get an error in the callback. It'll help us troubleshoot any problems we encounter down the line.

lib/services/pceMetaService/getExcluded.js
creindle: That's cool!

lib/services/pceMetaService/getExcluded.js
creindle: Good catch!! :O

lib/services/pceMetaService/getExcluded.js
creindle: Just to make sure.. do you mean:
1) `function getExcluded(location, type, context, callback)` - This gives me an error
2) `function getExcluded([location, type], context, callback)` - This gives me an error
3) `function getExcluded(locationType, context, callback)`

?

creindle: Sorry if it seems like a silly question... just though #2 would work! But it didn't with the comparison tests, and only #3 works!

lib/services/pceMetaService/getExcluded.js
```diff
@@ -0,0 +1,53 @@
...
+ * @summary returns a list of excluded PCE entries
+ * @param  {Array} [location, type] - (location) entered from user for specific location, (type) select to search from specific type
+ * @param  {Object} context - the session data
+ * @param  {Function} callback - the callback function used to process results
+ * @return {Array} res - Single array returning a list of excluded PCE entries
+ *
+ * @memberof PCEMetaService
+ * @instance
+ */
+function getExcluded(locationType, context, callback) {
```
mfuroyama: Sorry if I wasn't clear on my original comment, but yeah...let's break this `Array` type parameter into two separate, explicitly-named parameters, so that the function signature is more obvious:
```Javascript
function getExcluded(location, type, context, callback) {
```
...and don't forget to modify the JSDoc comments accordingly

lib/broker/models/orwpce-get-excluded.js
```diff
@@ -0,0 +1,56 @@
...
+                            + '1 ORWPCE EXCLUDE IMMUNIZATIONS\n'
+                            + '2 ORWPCE EXCLUDE SKIN TESTS\n'
+                            + '3 ORWPCE EXCLUDE PATIENT ED\n'
+                            + '4 ORWPCE EXCLUDE HEALTH FACTORS\n'
+                            + '5 ORWPCE EXCLUDE EXAMS',
+            example: '3',
+        }],
+
+        invokeRPC: (rpcArgs, session, callback) => {
+            pceMetaService.getExcluded(rpcArgs, session, callback);
```
mfuroyama: So you'll need to separate out `rpcArgs` here and pass in the appropriate number of arguments after you tweak the service function signature

lib/services/tests/getExcluded-spec.js
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env node
+
+'use strict';
+
+const modelFunction = require('../pceMetaService');
+const serviceTestUtils = require('./serviceTestUtils');
```
mfuroyama: ...then tweak the tests accordingly.  Thanks!!

lib/services/pceMetaService/getExcluded.js
creindle: Omg... I was overthinking this!

creindle: Ok! I got it to work :D

lib/broker/models/orwpce-get-excluded.js
creindle: This slipped my mind that invokeRPC isn't the direct call to the getExcluded function! I can change the parameters between the two lines :O

PR #59, opened 2017-11-03
VAM-98: ORWPCE GET EXCLUDED

creindle:


lib/services/pceMetaService/getExcluded.js
```diff
@@ -4,20 +4,19 @@ const ParameterService = require('../parameterService');
...
-function getExcluded(locationType, context, callback) {
+function getExcluded(location, type, context, callback) {
     let usrInputtedLoc;
-    const loc = locationType[0] || '';
-    const type = locationType[1] || '';
-    if (loc !== context.locationId[context.locationId.length - 1]) {
-        usrInputtedLoc = context.locationId.slice(0, context.locationId.length - 1) + loc;
+    if (location !== context.locationId[context.locationId.length - 1]) {
+        usrInputtedLoc = context.locationId.slice(0, context.locationId.length - 1) + location;
     }
```
mfuroyama: Cool! One last thing to address: Let's add some guard code or modify your logic here to ensure that `usrInputtedLoc` is never passed in to the `entArr` array below as `undefined`, which will happen if the following conditional fails:
```Javascript
if (location !== context.locationId[context.locationId.length - 1])
```

lib/broker/models/orwpce-get-excluded.js
```diff
@@ -41,7 +41,7 @@ const RPC_EMULATOR_MODEL = [
         }],
```
kekuso: Hmm weird, I can't pin this comment anywhere else but here...
Anyways, you might want to list the five used parameters in the 'parameters' property of this model.
@mfuroyama correct me if I'm wrong!

lib/services/pceMetaService/getExcluded.js
creindle: I believe I pushed up the changes to this... Weird... Let me check again.

lib/broker/models/orwpce-get-excluded.js
creindle: Ok! I'll see where I can add those in!

mfuroyama: @kekuso Yes, excellent suggestion!

mfuroyama: So, @creindle Let's add the 5 parameter names:

    ORWPCE EXCLUDE IMMUNIZATIONS
    ORWPCE EXCLUDE SKIN TESTS
    ORWPCE EXCLUDE PATIENT ED
    ORWPCE EXCLUDE HEALTH FACTORS
    ORWPCE EXCLUDE EXAMS

to the `parameters` array field in the RPC Model object

lib/services/pceMetaService/getExcluded.js
mfuroyama: I'm sorry, I don't think I was being clear here. Notice that `usrInputtedLoc` is declared without assignment on line 17:
```Javascript
let usrInputtedLoc;
```

So on line 20, if `loc` is, in fact, equivalent to the location id, then `usrInputtedLoc` is never assigned to anything...in essence, it's still `undefined`.

On line 29, you're using `usrInputtedLoc`, which may or may not be `undefined`, in your `entArr` array. Does that make sense?

A better approach would be to define the `entArr` first, without `usrInputtedLoc`. Then if the conditional passes, push the value of `usrInputtedLoc` in:
```Javascript
    const entArr = [
        context.userId,
        context.stationId,
        context.serviceId,
        context.divisionId,
        context.domainId,
    ];
    if (loc !== context.locationId[context.locationId.length - 1]) {
        entArr.push(context.locationId.slice(0, context.locationId.length - 1) + loc);
    }
```

Hopefully that clarifies things. If you're still confused or on the fence about what I'm asking for, let's discuss...

creindle: Ooooooooh, yes that's much more clear on what you meant. I apologize on that!

creindle: Also! That is a much better approach on this!

PR #61, opened 2017-11-04
VAM-126: ORWU PARAM

creindle:
- When `ParameterService.withContext(context).get(paramLabel)` in service file does not find entry from db, it jumps into catch and gives back error "Could not find definition for [param]".


lib/broker/models/orwu-param.js
```diff
@@ -0,0 +1,40 @@
...
+        inputParameters: [{
+            id: 'APARAM',
+            parameterType: 'LITERAL',
+            required: true,
+            description: 'The parameter to call, and return value from.',
+            example: 'ORWOR AUTO CLOSE PT MSG',
+        }],
+
+        invokeRPC: (rpcArgs, session, callback) => {
+            orderMetaService.orwuParam(rpcArgs, session, callback);
```
mfuroyama: Remember that `orwuParam` takes a `String` as a first parameter. `rpcArgs` here is always an array. Let's check/extract the argument prior to calling the service function

lib/broker/models/orwu-param.js
```diff
@@ -0,0 +1,40 @@
...
+        },
+        formatResult: (res) => {
```
mfuroyama: Truth be told, we don't need the `formatResult` function here

lib/services/orderMetaService/orwuParam.js
```diff
@@ -0,0 +1,33 @@
...
+ * @instance
+ */
+function orwuParam(paramLabel, context, callback) {
+    ParameterService.withContext(context).get(paramLabel)
+        .then((res) => {
+            if (_.isUndefined(res)) {
+                callback(null, '');
+                return;
+            }
+            callback(null, _.toNumber(res[0].value));
```
mfuroyama: Let's add a better guard on the `res` variable. We want to ensure that when we access `res[0].value` below, we won't bomb with an unhandled exception.

For example, consider the case where `res`, for whatever reason, comes back as an empty array. `res[0]` will return `undefined`, and attempting to access `undefined.value` will generate an exception.

lib/services/orderMetaService/orwuParam.js
```diff
@@ -0,0 +1,33 @@
...
+ * @param  {Function} callback Function called on process completion, with standard (err, result) signature.
+ * @return {[type]}            [description]
+ *
+ * @memberof OrderMetaService
+ * @instance
+ */
+function orwuParam(paramLabel, context, callback) {
+    ParameterService.withContext(context).get(paramLabel)
+        .then((res) => {
+            if (_.isUndefined(res)) {
```
mfuroyama: Also, consider using either a ternary or a lodash function here to tighten up the code here:

### Ternary
```Javascript
const result = dataGuardPasses(res) ? res[0].value : '';
callback(null, result);
```

### Lodash
```Javascript
const result = _.get(_.first(res), 'value', '');
callback(null, result);
```

lib/services/tests/orwuParam-spec.js
```diff
@@ -0,0 +1,26 @@
...
+
+describe('orwuParam asynchronous specs', () => {
+    let context;
+    beforeAll(() => {
+        context = serviceTestUtils.beforeAllScript(modelFunction).context;
+    });
+
+    it('numeric param should return parameter value or empty string', (done) => {
+        modelFunction.orwuParam(0, context, (err, res) => {
+            if (err) console.log(err);
```
mfuroyama: Let's make sure this test actually tests something! :)

lib/broker/models/orwu-param.js
creindle: That's so interesting! The comparison tests didn't catch that need for change at all!

lib/broker/models/orwu-param.js
creindle: :O

lib/services/orderMetaService/orwuParam.js
creindle: Ah! I keep on forgetting these pretty functions! @.@

lib/services/tests/orwuParam-spec.js
creindle: Yes, my apologizes.

lib/broker/models/orwu-param.js
```diff
@@ -0,0 +1,37 @@
...
+const RPC_EMULATOR_MODEL = [
+    {
+        id: 'ORWU_PARAM',
+        name: 'ORWU PARAM',
+        function: 'PARAM^ORWU',
+        parameters: [],
+        files: [],
+        domain: 'ORDER',
+        labels: ['META', 'MULTI-PARAM'],
+        services: [],
```
mfuroyama: Don't forget the `orderMetaService` here...

creindle: NOooooooo!!!! I added it to the other VAM, but I forgot thissss one!!! :O :O

PR #62, opened 2017-11-04
VAM-16: XUS GET USER INFO RPC/Service Implementation

mfuroyama:
This branch includes the following modules and tests:
### Services
* userMetaService
    * getUserInfo
* systemMetaService
    * getKernelSystemParameters

### Models
* XUS GET USER INFO


lib/broker/models/xus-get-user-info.js
```diff
@@ -0,0 +1,76 @@
...
+
+    invokeRPC: (rpcArgs, session, callback) => {
+        if (!session || !session.userId || !session.stationId) {
+            return callback(new Error('Invalid session context'));
+        }
+        userService.getUserInfo(session, callback);
+    },
+    formatResult: (res) => {
+        if (!_.isObject(res)) {
+            return new Error('Invalid response: Respons must be an object');
```
creindle: `return new Error('Invalid response: **Response** must be an object');` Sorry! Just caught my eye ;D

creindle: Yay it did show up! Also, sorry, I don't know if I should've just made a commit... but that's not how these work correct? We don't write commits on other pull requests correct?

kekuso: Christie = designated spell checker :D

mfuroyama: Thanks for catching this!

PR #64, opened 2017-11-06
VAM-72: ORWDXC ON

creindle:
- [X] Specific Error `Could not find definition for ORK SYSTEM ENABLE_DISABLE` due to special character search within Parameter Service resolved
Note: Original search has `ORK SYSTEM ENABLE/DISABLE` (p53_orwdxcOn.js)
- [X] All files created
- [X] Passes comparison tests 
Note: Must disable unrelated model `orwpt16-pscnvt` as it has no specs to test


lib/services/cprsMetaService/orwdxcOn.js
```diff
@@ -0,0 +1,27 @@
+'use strict';
```
mfuroyama: Ooh, I know this is weird, but could you move the service function over to `cprsConfigService` instead of creating a new `cprsMetaService`?  Basically this entails:
* [ ] Moving this file over to the `cprsConfigService` directory
* [ ] Changing the `@memberof` tag in the JSDoc comment
* [ ] Delete all things `cprsMetaService`

Thanks!

lib/services/cprsMetaService/orwdxcOn.js
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * @summary Returns E if order checking enabled, otherwise D.
+ * @param  {Object}   context  Session context information associated with the service call.
+ * @param  {Function} callback Function called on process completion, with standard (err, result) signature.
+ * @return {String} Either 'E' or 'D' representing order checking enabled or disabled
+ *
+ * @memberof CprsMetaService
+ * @instance
+ */
+function orwdxcOn(context, callback) {
```
mfuroyama: Also, let's consider renaming the service function. `orwdxcOn` has the lingering funk of RPCs.  Remember, the service function APIs are meant to (eventually) be used by external developers (like folks writing ExpressJS services, etc.)

Looking at the function, it looks like the purpose of this function is to see if order checking is enabled. So, maybe a better name for this function would be `isOrderCheckingEnabled`

lib/services/parameterService.js
```diff
@@ -226,8 +226,11 @@ class ParameterService extends BaseService {
         const service = this;
         return new Promise((resolve, reject) => {
             service.getDefinition(name, context).then((definition) => {
+                // Convert special characters to underscores because for some reason, VISTA stores Parameter Definitions
+                // with special characters, but then stores Parameters with underscores.
+                const parameterName = name.replace(/[/]/g, '_');
                 const findOptions = {
-                    query: service.buildParameterQuery(name, definition, options),
+                    query: service.buildParameterQuery(parameterName, definition, options),
```
mfuroyama: Who did this!? What a hack... ☹️

lib/services/cprsMetaService/orwdxcOn.js
creindle: Oooooh, I can do that, but the placement of the file creations were automatic. Should the locations of their creations change?

lib/services/cprsMetaService/orwdxcOn.js
creindle: HAhahahahahaha, ok! I'll change the function name to that!

lib/services/cprsMetaService/orwdxcOn.js
mfuroyama: Hmm...actually you raise a good point. On second thought, could you do me a huge favor? Could you move the stuff in `cprsConfigService` over into `cprsMetaService`? So, that would entail:
* [ ] Move the files in the `cprsConfigService` directory into `cprsMetaService`
* [ ] Change the @memberof tags in the files from `cprsConfigService` to point to `cprsMetaService`

Thanks!

mfuroyama: ...and then (sorry I wasn't clear about this), let's change the actual name of the service function module to match the function name, so:
`isOrderCheckingEnabled.js`

creindle: Yes, I was wondering if I should do this too because we would understand the reference to the nonClinical version of these RPCs!

lib/services/tests/orwdxcOn-spec.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
```
mfuroyama: Yeah, and here too...let's change the spec file name to match the service function module name

lib/services/cprsMetaService/orwdxcOn.js
creindle: So... I would change ***just*** the service file name?

PR #65, opened 2017-11-06
Vam 124 orwu haskey

kekuso:
punted! (see fileManService/getSecuritySpecs)


lib/services/fileManService/getSecuritySpecs.js
```diff
@@ -0,0 +1,25 @@
...
+ * @example
+ *  '63'
+ *
+ * @return Empty string for now...
+ *
+ * @memberOf  FileManService
+ * @instance
+ */
+function getSecuritySpecs(keyName, userId) {
+    // punt!
```
mfuroyama: 🏈  👍

PR #66, opened 2017-11-06
VAM-74: ORWDXM MSTYLE

creindle:
- [X] Created all files
- [X] Passed comparison tests
- [X] Moved files out of  `cprsConfigService` to `cprsMetaService` and adjusted file referencing and JSDocs accordingly for those files


lib/broker/models/orwpt16-pscnvt.js
```diff
@@ -1,31 +1,31 @@
-#!/usr/bin/env node
```
mfuroyama: Let's uncomment this model, please! 😄 

The way to fix the errors in the comparison is to pull the latest master branch from the metaVDP repo.

creindle: Ah! I apologize, the other branch should have this file uncommented as I pulled from the `metaVDP` repo as well. Let me check.

lib/services/cprsMetaService/orwdxmMstyle.js
```diff
@@ -0,0 +1,28 @@
+'use strict';
```
mfuroyama: Same thing here: Let's change the name of the service function module to match the name of the function: `menuStyle.js`

lib/services/tests/orwdxmMstyle-spec.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
```
mfuroyama: ...and probably a good idea to change the name of the spec file to match the service function module name..

PR #67, opened 2017-11-07
Performance test fixes

mfuroyama:
Out of band from what we're doing, but while I was going through the practice exercise of reorganizing our metaPureJS repo, I noticed that there were some defects in the performance/benchmark test suites. This branch fixes them:

* The native, RPC Broker socket-based test timing numbers were off, due to the asynchronous nature of the runs and the fact that the stopwatch events weren't happening as close to the wire as possible.
* The RPCs need to be run in strict serial order to keep the tests consistent, which they were not. This was fixed by using the async.eachSerial function to ensure that only one RPC was run at a time.
* The raw RPC message string translator function was broken, caused by an incorrect argument string length being passed in. Fixed this as well.

Here's a sample report with much better looking numbers: https://gist.github.com/mfuroyama/b2cf5ed50fad61912bfced54e08141d4


kekuso:
Awesome! but whoa, why is GMV-CONVERT-DATE taking so long?

PR #69, opened 2017-11-07
VAM-44: ORWCIRN AUTORDV

creindle:
- [X] All files created
- [X] Passed comparison tests


lib/services/orderMetaService/remotePatientQueryAuto.js
```diff
@@ -0,0 +1,34 @@
...
+ * @summary Get parameter value for ORWP CIRN AUTOMATIC
+ * @param  {Array}   usrInput  Determines to retrieve parameter value or not
+ * @param  {Object}   context  Session context information associated with the service call.
+ * @param  {Function} callback Function called on process completion, with standard (err, result) signature.
+ * @return {Number}  Parameter determines if remote patient queries are done automatically for all sites.
+ *
+ * @memberof OrderMetaService
+ * @instance
+ */
+function remotePatientQueryAuto(usrInput, context, callback) {
```
mfuroyama: Hmm, let's convert the expected type of `usrInput` from `Array` to something like `Boolean` or something, and let's rename the parameter. Remember, we want to be as explicit as possible when creating our service function signatures.

lib/services/orderMetaService/remotePatientQueryAuto.js
```diff
@@ -0,0 +1,34 @@
...
+    /* Main implementation */
+    if (!usrInput || !_.isArray(usrInput)) {
+        callback(Error);
```
mfuroyama: Instead of returning the `Error` class definition, let's return an `Error` *instance* instead:
```Javascript
callback(new Error('Some spiffy error message describing the fault'));
```

lib/services/orderMetaService/remotePatientQueryAuto.js
```diff
@@ -0,0 +1,34 @@
...
+        return;
+    }
+    ParameterService.withContext(context).get('ORWRP CIRN AUTOMATIC', { entity: 'ALL' })
+        .then((res) => {
+            const result = usrInput.length === 0 ? 0 : res[0].value;
```
mfuroyama: Consider adding a guard to `res` here. Remember that if `res` is an empty array, `res[0]` will evaluate to `undefined`, which will cause the `value` field access to hard-fail.

lib/services/orderMetaService/remotePatientQueryAuto.js
creindle: When you mean convert it to `Boolean`, are you talking about the !! operator? Why change it to `Boolean` specifically? Also, I didn't exactly grab what the parameter usually was for the service function from reading the Rambler or the http://vistadataproject.info/.

lib/services/orderMetaService/remotePatientQueryAuto.js
```diff
@@ -0,0 +1,29 @@
...
+ * @return {Number}  Parameter determines if remote patient queries are done automatically for all sites.
+ *
+ * @memberof OrderMetaService
+ * @instance
+ */
+function remotePatientQueryAuto(context, callback) {
+    /* Main implementation */
+    ParameterService.withContext(context).get('ORWRP CIRN AUTOMATIC', { entity: 'ALL' })
+        .then((res) => {
+            const resultValue = res[0] === undefined ? 0 : res[0].value;
```
mfuroyama: Consider using the following pattern here: 
```Javascript
const resultValue = _.get(res[0], 'value', 0);
```

As it is now, if `res` is an array, but the first value doesn't contain a `value` field, `resultValue` will contain `undefined`. With the assignment pattern I suggested, you catch *all* the error cases.

creindle: Ahhhh! Ok, yes this would be muchhh better

PR #70, opened 2017-11-07
Vam 26 gmv param

sidjtd:
Context.name (line 84) needs to be re-factored to utilize User Service but other than that, hopefully all is good to go!!!!

(removed a lot of returns in front of callbacks to make them one liners... please lmk if that was not a good idea, reverting it back to all having 'return callbacks()' is too easy!


lib/broker/models/gmv-parameter.js
```diff
@@ -0,0 +1,107 @@
...
+            'GMV ALLOW USER TEMPLATES',
+            'GMV DEFAULT VALUES',
+            'GMV TEMPLATE',
+            'GMV TEMPLATE DEFAULT',
+            'GMV USER DEFAULTS',
+        ],
+        files: ['8989.3', '8989.51', '8989.518'],
+        domain: 'CPRS',
+        labels: ['META', 'VITAL', 'CHANGE'],
+        services: [],
```
mfuroyama: Mr. Pedantic reluctantly mumbles: let's add 'cprsMetaService' here... 😄

lib/services/cprsMetaService/gmvParameter.js
```diff
@@ -0,0 +1,163 @@
...
+/**
+ * @summary Sets and retrieves parameter values used by the graphical user interface.
+ * @param {String} routineArg Routine tag line to call
+ * @param {String} entity The entity value to use
+ * @param {String} gmvParamString Parameter string used to call parameter service
+ * @param {String} instance The instance to use.
+ * @param {String} value The value assigned to a parameter, stored in FILE 8989.5
+ * @param  {Object}   context  Session context information associated with the service call.
+ * @param  {Function} callback Function called on process completion, with standard (err, result) signature.
+ * @return {[type]}            [description]
```
mfuroyama: I'm guessing but I think the return type here should be `Array`? Really minor but could we add this in? Thanks!

lib/services/cprsMetaService/gmvParameter.js
```diff
@@ -0,0 +1,163 @@
...
+                return callback(null, output);
+            })
+            .catch((err) => {
+                console.log(err);
+                callback(err);
+            });
+    }
+
+    // Argument 'ENTVAL' for entity value information
+    if (routineArg === 'ENTVAL') {
```
mfuroyama: Really minor, but consider changing this to an `else if`.

lib/services/cprsMetaService/gmvParameter.js
```diff
@@ -0,0 +1,163 @@
...
+                });
+        } else if (entityArg === 'DIV') return { service: context.division.label, key: context.division.label };
+        else if (entityArg.match(/(\d+);?\D{3}?\((\d+)/)) {
+            const matched = entityArg.match(/(\d+);?\D{3}?\((\d+)/);
+            const fieldId = matched[1];
+            const fileNumber = matched[2];
+            const fileName = fileman.getFileLabelFromId(fileNumber).replace(/\s/g, '');
+
+            this.find(fileName, {
+                query: { id: `urn:vista:${context.stationId}:${fileNumber}:${fieldId}` },
```
mfuroyama: Not a big deal, and you don't have to, but you might want to consider using `serviceUtils.createURNFromParams`...totally up to you

lib/services/fileManService/getFieldFromFile.js
```diff
@@ -2,11 +2,29 @@
 
 const vdmModels = require('../fileManService/vdmNonClinicalModel').vdmModel;
 const _ = require('lodash');
-
+// const util = require
```
mfuroyama: Ahh, you can probably delete this line out

lib/broker/models/gmv-parameter.js
sidjtd: DOH!

lib/services/cprsMetaService/gmvParameter.js
sidjtd: Noooo these metadata entriessssss *cries*

PR #71, opened 2017-11-08
VAM-78: ORWDXVB3 SWPANEL

creindle:
- [X] All files created
- [X] Passed comparison tests


lib/services/orderMetaService/diagnosticComponentPanelLoc.js
```diff
@@ -0,0 +1,30 @@
+'use strict';
+
+const ParameterService = require('../parameterService');
+const _ = require('lodash');
+
+/**
+ * @summary This RPC checks the value of the parameter OR VBECS DIAGNOSTIC PANEL 1ST to determine the location of the Diagnostic and Component panels on the VBECS order dialog.
+ * @param  {[type]}   args     [description]
```
mfuroyama: Let's remove this `@param` tag for `args` since: 1) It's not applicable, and 2) because it would break the service doc generator...

creindle: Noooooo! I missed thisssss NOooooo... T.T

lib/broker/models/orwdxvb3-swpanel.js
```diff
@@ -0,0 +1,34 @@
...
+        labels: ['META'],
+        services: ['orderMetaService'],
+        mvdm: {},
+        rpcType: 'READ',
+        complexity: 'LOW',
+        description: 'This RPC checks the value of the parameter OR VBECS DIAGNOSTIC PANEL 1ST to determine the location of the Diagnostic and Component panels on the VBECS order dialog.',
+        returnType: 'SINGLE VALUE',
+        inputParameters: [],
+        invokeRPC: (rpcArgs, session, callback) => {
+            if (rpcArgs.length === 0) {
```
kekuso: Don't think you need this check since you won't be passing the rpcArgs param to the service

lib/services/tests/diagnosticComponentPanelLoc-spec.js
```diff
@@ -0,0 +1,21 @@
...
+    beforeAll(() => {
+        context = serviceTestUtils.beforeAllScript(modelFunction).context;
+    });
+
+
+    it('Expect res to be defined and number', (done) => {
+        modelFunction.diagnosticComponentPanelLoc(context, (err, res) => {
+            if (err) console.log(err);
+            expect(res).toBeDefined();
+            expect(res).toEqual(jasmine.any(Number));
```
kekuso: Pretty sure you can test specifically for 0 or 1 instead of Number
`expect(res === 0 || res === 1).toEqual(true)`

lib/broker/models/orwdxvb3-swpanel.js
creindle: screen shot 2017-11-08 at 7 51 00 am

creindle: It passed without it! But I saw this in the test and figured it needed a check for it @.@

creindle: Weird... cause the last RPC had the exact same setup for the specs

lib/services/tests/diagnosticComponentPanelLoc-spec.js
creindle: Oooh... I didn't realize that the only numbers that would come back are 0 and 1 😱

creindle: I'll add this though :D

PR #73, opened 2017-11-08
VAM-114: ORWPS REASON

creindle:
- [X] Created all files
- [X] Passes comparison tests


lib/broker/models/orwps-reason.js
```diff
@@ -0,0 +1,40 @@
...
+        rpcType: 'READ',
+        complexity: 'LOW',
+        description: 'Returns list of Statement/Reasons for Non-VA medication orders.',
+        returnType: 'ARRAY',
+        inputParameters: [],
+        invokeRPC: (rpcArgs, session, callback) => {
+            medMetaService.nonVAMedOrderReasons(session, callback);
+        },
+        formatResult: (res) => {
+            if (_.isUndefined(res) || !_.isArray(res)) return new Error('res is either undefined or not array');
```
mfuroyama: Let's add some kind of guard to ensure that `res`, does in fact, have 4 values in the array. Any short of that, and the code below will bomb out with an unhandled exception...

creindle: Ah yes! That's a good suggestion!

PR #76, opened 2017-11-09
VAM-93: ORWPCE AUTO VISIT TYPE SELECT

creindle:


lib/broker/models/orwpce-auto-visit-type-select.js
```diff
@@ -0,0 +1,36 @@
...
+        inputParameters: [{
+            id: 'LOC',
+            parameterType: 'LITERAL',
+            required: false,
+            description: 'location ID of visit',
+            example: '1',
+        }],
+
+        invokeRPC: (rpcArgs, session, callback) => {
+            pceMetaService.autoVisitType(rpcArgs, session, callback);
```
mfuroyama: Hmm, so remember that in the `invokeRPC` parameter will **always** be an array. And remember that we want to avoid passing/using arrays to our service functions...

lib/services/pceMetaService/autoVisitType.js
```diff
@@ -0,0 +1,38 @@
...
+ * @memberof PceMetaService
+ * @instance
+ */
+function autoVisitType(location, context, callback) {
+    /* Main implementation */
+    if (location === null) {
+        callback(null, new Error('Argument is null'));
+        return;
+    }
+    const loc = location[0] || '';
```
mfuroyama: 🤔  Interesting...so, in the JSDoc, you've declared that the `location` parameter should be a **Number**, yet here, you're treating it like an array...Hmm...also remember that we should try to avoid passing arrays into our service functions. 🤔  See my first comment for a solution to this dilemma... 😄

lib/services/pceMetaService/autoVisitType.js
```diff
@@ -0,0 +1,38 @@
...
+    const entArr = ['USR', 'SRV', 'DIV', 'SYS'];
+    if (loc !== context.locationId[context.locationId.length - 1]) {
```
mfuroyama: Could you explain to me what this is doing exactly? I'm wondering if there's a different pattern we can use here...

lib/broker/models/orwpce-auto-visit-type-select.js
creindle: UGHHH noooo!

lib/services/pceMetaService/autoVisitType.js
creindle: Ah.. that's a good point though!

lib/services/pceMetaService/autoVisitType.js
creindle: This checks if the inputted location matches with the context location, and if it doesn't, then replace the location with the inputted location!

lib/broker/models/orwpce-auto-visit-type-select.js
creindle: I fixed it!

lib/services/pceMetaService/autoVisitType.js
mfuroyama: Ahh, okay, I see your intent, but what I wanted you to explain was the detail:

From what I can tell, it looks like you're grabbing the **last** character out of the `locationId` string, then using that as the basis of your equality comparison. So, what happens in the case where the `locationId` string is set to, say, `"urn:vista:6100:44:21"`? An equality check would pass, even though it should fail.

This gets back to the idea of writing your code not just for the specific situation and data at hand, but to handle the general intent and expectations. Only make assumptions about the data you're working with when those assumptions are **guaranteed** and solid.  Remember, we're hard-coding the context object information to provide a **sample** data set to work with...the data in the context object will definitely be different in operations, so you need your code to work within those constraints.

For the particular problem presented here, let's rethink your case comparison logic and try a different, more general approach. Here's a hint: consider using the `serviceUtils.getIenFromId()` utility function.

creindle: That's a very good point!

lib/services/pceMetaService/autoVisitType.js
```diff
@@ -0,0 +1,41 @@
...
+    }
+    const loc = location || '';
+    const entArr = ['USR', 'SRV', 'DIV', 'SYS'];
+    if (serviceUtils.getIenFromId(context.locationId) !== loc && _.isNumber(loc)) {
+        const locUrn = serviceUtils.createURNFromParams('44', loc, context.stationId);
+        entArr.push(locUrn);
+    }
+    ParameterService.withContext(context).get('ORWPCE DISABLE AUTO VISIT TYPE', { entity: entArr })
+        .then((res) => {
+            const resValue = res.length === 0 ? 1 : 0;
```
sidjtd: `const resValue = res.length === 0 ? 1 : 0;
`
Hmmm how about something like... 

`const resValue = +_.isEmpty(res);`

So giggity.

creindle: That's so cool!

PR #79, opened 2017-11-09
Vam 79 ask pce

sidjtd:
Simplified from original implementation that called in user data, now just does file 200 mongo query or defaults to context data instead


lib/services/pceMetaService/askPCE.js
```diff
@@ -0,0 +1,48 @@
...
+    // If incoming argument is a valid number, use that as Id arg. Else, use current users Id
+    const validNumber = (_.isNumber(parseInt(userId, 10)));
+    const contextId = serviceUtils.getIenFromId(context.userId);
+    const idToCheck = (!validNumber) ? contextId : userId;
+
+    const urnObj = {
+        fileType: 200, ien: idToCheck, isNational: !context.stationId, station: context.stationId,
+    };
+
+    const urnString = serviceUtils.createURNFromIdObject(urnObj);
```
mfuroyama: Consider using `serviceUtils.createURNFromParams` here. The difference is subtle, but the benefits are:
1. You don't need to craft an object to pass in,
2. You don't need to bother with `isNational`, the function takes care of that for you.
so:
```Javascript
const urnString = serviceUtils.createURNFromParams(200, idToCheck, context.stationId);
```
...easy peasy!

lib/services/pceMetaService/askPCE.js
```diff
@@ -0,0 +1,48 @@
...
+    const stationId = context.locationStationId;
+
+    this.find('NewPerson', {
+        query: { id: urnString },
+        projection: serviceUtils.createProjection(['serviceSection']),
+
+    }).then((userQuery) => {
+        const serviceId = (!validNumber) ? context.serviceId : userQuery[0].serviceSection.id;
```
mfuroyama: You might want to consider adding a guard on the `userQuery` array, just in case the query comes back with an empty array...

lib/services/pceMetaService/askPCE.js
sidjtd: so much cleaner

PR #80, opened 2017-11-13
VAM-141: TIU TEMPLATE GET DEFAULTS

creindle:
- [X] Created all files and deleted unnecessary ones
- [X] Passed comparison tests


emulation/models/tiu-template-get-defaults.js
```diff
@@ -0,0 +1,47 @@
...
+                callback(null, '');
+                return;
+            }
+            if (_.find(rpcArgs, Number) ||
+                _.find(rpcArgs, String) ||
+                _.find(rpcArgs, null)) {
+                callback(new Error('Incorrect argument format'));
+                return;
+            }
+            ParameterService.withContext(session).get('TIU DEFAULT TEMPLATES', { entity: 'USR', format: 'Q' })
```
creindle: I'm not 100% sure how to implement the `format: 'Q'` portion. Is this correct?

emulation/models/tiu-template-get-defaults.js
```diff
@@ -0,0 +1,47 @@
...
+        complexity: 'LOW',
+        description: 'Returns Default Template Settings.',
+        returnType: 'SINGLE VALUE',
+        inputParameters: [],
+        invokeRPC: (rpcArgs, session, callback) => {
+            if (_.isEmpty(rpcArgs)) {
+                callback(null, '');
+                return;
+            }
+            if (_.find(rpcArgs, Number) ||
```
creindle: The `nonClinicalRPC` spec of this RPC (p127-spec.js) tested for these various arguments. Is it necessary to include them if they don't change the output of the function? I would believe not as it didn't change the success of the comparison tests.

services/documentMetaService.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
```
creindle: This file was created after the `npm run make` command, should I delete this one since it isn't used?

emulation/models/tests/tiu-template-get-defaults-spec.js
```diff
@@ -0,0 +1,40 @@
...
+const unitTestUtils = require('./unitTestUtils');
+const serviceTestUtils = require('../../../services/tests/serviceTestUtils');
+
+describe('TIUTEMPLATEGETDEFAULTS asynchronous specs', () => {
+    let rpc;
+    let context;
+
+    beforeAll(() => {
+        const res = unitTestUtils.beforeAllScript(rpcEmulatorModel);
+        const serviceRes = serviceTestUtils.beforeAllScript();
```
creindle: Not too sure if this is how we should implement this version to test the function in the model. Is this the best way? Should I have included the `serviceTestUtils` even though it doesn't have its own service function?

emulation/models/tiu-template-get-defaults.js
```diff
@@ -0,0 +1,47 @@
...
+
+    {
+        id: 'TIU_TEMPLATE_GET_DEFAULTS',
+        name: 'TIU TEMPLATE GET DEFAULTS',
+        function: 'GETDFLT^TIUSRVT2',
+        parameters: ['TIU DEFAULT TEMPLATES'],
+        files: [],
+        domain: 'DOCUMENT',
+        labels: ['META'],
+        services: [],
```
creindle: Since this has no service function implementation, would I still need to add the string 'documentMetaService' here?

creindle:
I just saw @kekuso 's pull request that including the rpc.run() method in unitTestUtils, so I'll wait for his to close then will pull.

emulation/models/tiu-template-get-defaults.js
mfuroyama: no, 'parameterService' is good enough

emulation/models/tiu-template-get-defaults.js
mfuroyama: Wait...could you explain to me what these tests are doing?

emulation/models/tiu-template-get-defaults.js
mfuroyama: You can remove the `format: 'Q'` from the ParameterService `get` call. That meant something in the old VDP domain, but it doesn't do anything here.

services/documentMetaService.js
mfuroyama: yes, delete please! Thanks!

emulation/models/tests/tiu-template-get-defaults-spec.js
mfuroyama: Hmm, yeah...we'll have to move some of the service test utility code back into the `beforeAllScript` function of the `models/tests/unitTestUtils.js` module...

emulation/models/tests/tiu-template-get-defaults-spec.js
```diff
@@ -0,0 +1,38 @@
...
+    let context;
+
+    beforeAll(() => {
+        const res = unitTestUtils.beforeAllScript(rpcEmulatorModel);
+        context = res.context;
+        rpc = res;
+    });
+
+    it('invoking invokeRPC with numeric arguments should return error', () => {
+        rpc.rpc.invokeRPC([1], context, (err, res) => {
```
creindle: I need to change these to `rpc.run` too to bring in Adams code!

PR #81, opened 2017-11-14
Vam 60 val num format

sidjtd:
Validate Numbers...


mfuroyama:
Hmm, after talking to Conor about service functions, I'm wondering if this would be better handled by moving the code out of service and back into the RPC model?

I think we need to come up with a good rule-of-thumb for what should go into a service and what should stay in the model. Perhaps, RPC emulation that is:

* Simple parameter only
* No MongoDB data access

sidjtd:
Makes sense, let me re-factor then!

PR #85, opened 2017-11-15
Vam-90 always checkout

sidjtd:
Only model and model-spec in this one


emulation/models/orwpce-always-checkout.js
```diff
@@ -0,0 +1,44 @@
...
+const RPC_EMULATOR_MODEL = [
+    {
+        id: 'ORWPCE_ALWAYS_CHECKOUT',
+        name: 'ORWPCE ALWAYS CHECKOUT',
+        function: 'DOCHKOUT^ORWPCE2',
+        parameters: ['ORWPCE DISABLE AUTO CHECKOUT'],
+        files: [],
+        domain: 'PCE',
+        labels: ['META'],
+        services: ['pceMetaService'],
```
mfuroyama: Let's change this to 'parameterService' since we've done away with the PCE meta service function

emulation/models/orwpce-always-checkout.js
```diff
@@ -0,0 +1,44 @@
...
+            example: '1',
+        }],
+
+        invokeRPC: (rpcArgs, session, callback) => {
+            if (_.isEmpty(session.userId)) return callback(null, '');
+            const entities = [session.stationId, session.serviceId, 'DIV', 'SYS'];
+
+            ParameterService.withContext(session).get('ORWPCE DISABLE AUTO CHECKOUT', { entity: entities })
+                .then((res) => {
+                    if (_.isEmpty(parseInt(res, 10)) || !res) return callback(null, 1);
```
mfuroyama: Consider consolidating this into a single `callback` call using a ternary to determine the appropriate output:
```Javascript
const result = (_.isEmpty(parseInt(res, 10)) || !res) ? 1: 0;
```

PR #86, opened 2017-11-15
Vam 91 pce anytime

sidjtd:
Another model & model-spec only implementation


emulation/models/orwpce-anytime.js
```diff
@@ -0,0 +1,38 @@
...
+const RPC_EMULATOR_MODEL = [
+    {
+        id: 'ORWPCE_ANYTIME',
+        name: 'ORWPCE ANYTIME',
+        function: 'ANYTIME^ORWPCE2',
+        parameters: ['ORWPCE ANYTIME ENCOUNTERS'],
+        files: [],
+        domain: 'PCE',
+        labels: ['META'],
+        services: ['pceMetaService'],
```
mfuroyama: Let's change this to 'parameterService' here, since we've done away with the PCE meta service function.

emulation/models/orwpce-anytime.js
```diff
@@ -0,0 +1,38 @@
...
+        returnType: 'SINGLE VALUE',
+        inputParameters: [],
+
+        invokeRPC: (rpcArgs, session, callback) => {
+            if (_.isEmpty(session.userId)) return callback(null, '');
+            const entities = [session.stationId, session.serviceId, 'DIV', 'SYS'];
+
+            ParameterService.withContext(session).get('ORWPCE ANYTIME ENCOUNTERS', { entity: entities })
+                .then((res) => {
+                    if (_.isEmpty(parseInt(res, 10)) || !res) return callback(null, 1);
```
mfuroyama: Same thing here: Consider consolidation to one `callback` call using a ternary to determine the output result.

emulation/models/orwpce-anytime.js
sidjtd: Got it

PR #87, opened 2017-11-15
VAM-18: ORDEA DEATEXT

creindle:
- [X] Created model and model spec file
- [X] Passed model unit tests
- [] Passed comparison tests 
Note: Session domainId does not match entity.id of OR DEA TEXT, therefore does not grab entry in MongoDB. DomainId is changed in model Jasmine spec file, which is why model unit tests pass, not comparison tests.


emulation/models/ordea-deatext.js
```diff
@@ -0,0 +1,46 @@
...
+        inputParameters: [],
+        invokeRPC: (rpcArgs, session, callback) => {
+            let splitResponse;
+            if (_.some(rpcArgs, item => (_.isNumber(item) || _.isString(item) || _.isNull(item)))) {
+                return callback(new Error('Invalid parameters'));
+            }
+
+            ParameterService.withContext(session).get('OR DEA TEXT', { entity: 'SYS' })
+                .then((res) => {
+                    if (_.isEmpty(res) || _.isUndefined(res[0].wordProcessingText)) {
```
sidjtd: `if (_.isEmpty(res) || _.isUndefined(res[0].wordProcessingText)) {`

This is personal pref so don't take it as a correction, maybe a bang (!) in front of 'res[0].wordProcessingText' may shorten it up, or combining the two into 1 call.  "_.isEmpty(res[0].wordProcessingText)". _.isEmpty

 I think you can do the 'splitResponse' variable declaration in 35 when it's first being used, instead of up in line 24. I appears that it'll be ok even if splitResponse is not at the top lexical scope since it just goes right into the callback...

creindle: That's a good point! I'll add those suggestions in though!

emulation/models/ordea-deatext.js
```diff
@@ -0,0 +1,46 @@
...
+        rpcType: 'READ',
+        complexity: 'MEDIUM',
+        returnType: 'SINGLE VALUE',
+        description: 'Returns DEA Mandated Signature Dialog Text when controlled substance order is selected to be signed.',
+        inputParameters: [],
+        invokeRPC: (rpcArgs, session, callback) => {
+            let splitResponse;
+            if (_.some(rpcArgs, item => (_.isNumber(item) || _.isString(item) || _.isNull(item)))) {
+                return callback(new Error('Invalid parameters'));
+            }
```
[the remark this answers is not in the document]
creindle: I would definitely agree that it may be unnecessary because the comparison tests still pass without them! But, I'll wait on Mike too!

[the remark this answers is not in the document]
mfuroyama: Actually, yeah...after looking at this, I agree with @sidjtd. Let's remove this check since, like we discussed this morning, the parameters should already be being checked by the dispatcher in the default validity checks.

emulation/models/ordea-deatext.js
```diff
@@ -0,0 +1,46 @@
...
+
+            ParameterService.withContext(session).get('OR DEA TEXT', { entity: 'SYS' })
+                .then((res) => {
+                    if (_.isEmpty(res) || _.isUndefined(res[0].wordProcessingText)) {
+                        callback(null, '');
+                        return;
+                    }
+                    splitResponse = _.split(res[0].wordProcessingText, '\n', 2);
```
mfuroyama: Could you explain why you added the line limit parameter value (2) to the `split` call?

emulation/models/ordea-deatext.js
```diff
@@ -0,0 +1,46 @@
...
+                    splitResponse = [splitResponse[0], _.trimEnd(splitResponse[1])];
```
mfuroyama: Hm...I'm trying to understand the intent here as well...could you help me out? 😄

emulation/models/ordea-deatext.js
creindle: I saw that in the example on the Lodash site. I wanted to format the output of `res[0].wordProcessingText`

emulation/models/ordea-deatext.js
creindle: So, after splitting the value of `res[0].wordProcessingText` into an array of length `2` based on the line break, the comparison tests would not pass unless I removed the white spaces at the end of the first sentence after the split.

emulation/models/ordea-deatext.js
creindle: I did see removing the third parameter made no difference in the output of the code though!

mfuroyama: I have a feeling that this might be another case of coding to a narrow, specific case vs. coding to handle a broader scope.

I cannot stress this enough: you *must* resist the urge to write code that conforms to a *specific* test data case. 😄  Keep the bigger picture/functional purpose in mind when you write your code...always be asking yourself: "What is the **true** purpose of this code?"

In this specific case, the purpose of this function should be to: "Get the word processing string from the result, split that string according to `\n` characters and return the lines in an array"

The intent I read from your code is: "Get the word processing string, split the string but **always and only** return the first two lines. The results should have **no more than 2 lines of text**". But what if, for example, the `wordProcessingText` in a different system collection had more than 2 lines? By making narrow assumptions based on a single set of sample result data, we inadvertently break the contract outlined by the original requirement.

emulation/models/ordea-deatext.js
mfuroyama: Here as well: I have a feeling that this might be another case of coding to a narrow, specific case vs. coding to handle a broader scope.

The next line strengthens the implied purpose of this code, which seems to be: "The results should have **no more than 2 lines of text**"

See my comment above for details

emulation/models/ordea-deatext.js
creindle: Thank you for the talk and explanation! I will definitely ask for clarification on the purpose of the RPCs in the future!

mfuroyama:
Hi hi, @creindle!! I finally got around to checking this out. There were a couple logic problems that were pretty easy to fix. For the record, please review the implemented changes, then you accept the pull request, I'll merge in.

PR #88, opened 2017-11-15
Vam 19 new reminder

sidjtd:
3rd ones the charm?


emulation/models/orqqpx-new-reminders-active.js
```diff
@@ -0,0 +1,37 @@
...
+const RPC_EMULATOR_MODEL = [
+    {
+        id: 'ORQQPX_NEW_REMINDERS_ACTIVE',
+        name: 'ORQQPX NEW REMINDERS ACTIVE',
+        function: 'NEWACTIV^ORQQPX',
+        parameters: ['PXRM GUI REMINDERS ACTIVE'],
+        files: [],
+        domain: 'ORDER',
+        labels: ['META'],
+        services: ['orderMetaService'],
```
mfuroyama: Let's change this to 'parameterService' here

emulation/models/orqqpx-new-reminders-active.js
```diff
@@ -0,0 +1,37 @@
...
+        description: 'Return 1 if Interactive Reminders are active, otherwise return 0.',
+        returnType: 'SINGLE VALUE',
+        inputParameters: [],
+        invokeRPC: (rpcArgs, session, callback) => {
+            if (_.isEmpty(session.userId)) return callback(null, '');
+            const entities = [session.stationId, session.serviceId, 'DIV', 'SYS'];
+
+            ParameterService.withContext(session).get('PXRM GUI REMINDERS ACTIVE', { entity: entities })
+                .then((res) => {
+                    if (_.isEmpty(parseInt(res, 10)) || !res) return callback(null, 1);
```
mfuroyama: Same here: Consider consolidating to one `callback` call using a ternary to determine the appropriate result value

services/orderMetaService/newReminder.js
```diff
@@ -0,0 +1,30 @@
+'use strict';
```
mfuroyama: Hmm...do we need this service function? Perhaps an artifact from the template generator? I don't see it being used in the RPC model `invokeRPC` anywhere...

services/tests/newReminder-spec.js
```diff
@@ -0,0 +1,21 @@
...
+    let context;
+    beforeAll(() => {
+        context = serviceTestUtils.beforeAllScript(modelFunction).context;
+    });
+
+
+    it('Empty arg returns .....', (done) => {
+        modelFunction.newReminder([], context, (err, res) => {
+            if (err) console.log(err);
+            // optionalTestFunction(res);
```
mfuroyama: It looks like this Jasmine spec isn't doing much... 😄  Plus, we might not need it anyway (see above comment)

PR #90, opened 2017-11-15
VAM-118: ORWTPD1 GETEAFL

creindle:
- [X] Created model and spec file 
- [X] Passes model tests and comparison tests


emulation/models/orwtpd-1-geteafl.js
```diff
@@ -0,0 +1,45 @@
...
+        complexity: 'LOW',
+        description: 'Get “future days limit” parameter setting for Enc appt selection.',
+        returnType: 'SINGLE VALUE',
+        inputParameters: [],
+        invokeRPC: (rpcArgs, session, callback) => {
+            if (_.isEmpty(rpcArgs)) {
+                callback(null, '');
+                return;
+            }
+            const entArr = [
```
mfuroyama: So, let's replace this with `['DIV', 'SYS', 'PKG', 'USR']` since the parameter service has extra code in here to take care of these cases.

Also of note, the `session` context object no longer has a `packageId` field, since the Parameter Service actually determines that field for you...it's in the MUMPS... :(

creindle: Ooooh!! Yes, I completely forgot that we can simply add those!

PR #92, opened 2017-11-16
Refactoring param services

creindle:


emulation/models/orwdxc-on.js
```diff
@@ -13,14 +13,21 @@ const RPC_EMULATOR_MODEL = [
...
         rpcType: 'READ',
         complexity: 'LOW',
         description: 'Returns E if order checking enabled, otherwise D.',
         returnType: 'SINGLE VALUE',
         inputParameters: [],
         invokeRPC: (rpcArgs, session, callback) => {
-            cprsMetaService.isOrderCheckingEnabled(session, callback);
+            ParameterService.withContext(session).get('ORK SYSTEM ENABLE/DISABLE', { entity: 'ALL' })
+                .then((res) => {
+                    const response = res[0].value ? res[0].value : '';
```
mfuroyama: Let's use the pattern you've been using in other implementations here:
```Javascript
const resultValue = _.get(res[0], 'value', '');
```

As it is now, if `res` is empty, this code will crash...

creindle: Oh no... I didn't think it would crash as `res[0].value` would give undefined. Using `_.get` is much better.

PR #93, opened 2017-11-16
Refactoring param services 

creindle:
- [X] Grabbed from latest master
- [X] Passes model tests
- [X] Passes comparison tests


emulation/models/orwch-ldfont.js
```diff
@@ -32,13 +32,20 @@ const RPC_EMULATOR_MODEL = [
...
         }],
-        invokeRPC: (args, session, callback) => {
+        invokeRPC: (rpcArgs, session, callback) => {
             if (_.isEmpty(session.userId)) {
                 callback(null, '');
                 return;
             }
-            // Get the font size via ParameterService
-            OrderMetaService.loadFont(session, callback);
+            ParameterService.withContext(rpcArgs[0]).get('ORWCH FONT SIZE', { entity: 'USR' })
```
mfuroyama: Let's change the value passed to `withContext` from `rpcArgs[0]` to `context`. It should look like this:
```Javascript
ParameterService.withContext(context).get('ORWCH FONT SIZE', { entity: 'USR' })
```

emulation/models/orwch-ldfont.js
```diff
@@ -32,13 +32,20 @@ const RPC_EMULATOR_MODEL = [
...
+                .then((result) => {
+                    const value = (_.isArray(result) && result.length > 0) ? result[0].value : result.value;
```
mfuroyama: Hmm, could you explain the intent of this line of code to me?

emulation/models/orwch-ldfont.js
```diff
@@ -32,13 +32,20 @@ const RPC_EMULATOR_MODEL = [
...
+                    callback(null, value);
+                })
+                .catch((err) => {
+                    console.log(err);
```
mfuroyama: Let's remove this `console.log` here too...

emulation/models/orwch-loadsiz.js
```diff
@@ -28,13 +28,23 @@ const RPC_EMULATOR_MODEL = [
...
             if (_.isEmpty(session.userId)) {
                 callback(null, '');
                 return;
             }
-            // Get font size via ParameterService
-            OrderMetaService.loadSize(args, session, callback);
+            ParameterService.withContext(session).get('ORWCH BOUNDS', { entity: 'USR', instance: rpcArgs[0] })
+                .then((res) => {
+                    if (_.isString(rpcArgs[0]) && _.has(res, 'value')) {
+                        const val = (_.isArray(res) && res.length > 0) ? res[0].value : res.value;
```
mfuroyama: This is a weird one as well, same as the previous line. Could you explain this one to me as well?

emulation/models/orwch-loadsiz.js
```diff
@@ -28,13 +28,23 @@ const RPC_EMULATOR_MODEL = [
...
+                        callback(null, val);
+                    }
+                    callback(null, '');
+                })
+                .catch((err) => {
+                    console.log(err);
```
mfuroyama: Let's delete the `console.log` here as well

emulation/models/orwch-ldfont.js
creindle: Oh no I thought I changed that! :O

emulation/models/orwch-loadsiz.js
creindle: I'm pretty sure I modified these...

emulation/models/orwch-ldfont.js
creindle: I didn't do the original implementation of this RPC, I transferred them over from the service function implementation. Here is the original pull request that utilized this: https://github.com/vistadataproject/VICSServer/pull/5/files

emulation/models/orwch-ldfont.js
creindle: I'll check again, but I don't think changing it to `context` worked for me.

emulation/models/orwch-ldfont.js
creindle: This was also another one I transferred over from the service function implementation. So I unfortunately did not test exactly what it was doing. Here is the pull request with the original code: https://github.com/vistadataproject/VICSServer/pull/8/files

emulation/models/orwch-loadsiz.js
creindle: This was another one I transferred over from the service function implementation, so unfortunately I did not test it. Here is the original implementation of this function: https://github.com/vistadataproject/VICSServer/pull/8/files

emulation/models/orwch-ldfont.js
creindle: So I'm not 100% what the original implementation was meant to do.

mfuroyama: Ahh, okay...yeah, the reason I asked was because this code doesn't make sense to me. `result` should always be an array, but notice that the second part of the ternary is returning a `result.value`, which will **always** be undefined.

Okay, so you didn't write the original, but you're going to fix it here!  :)  So, knowing that `result` should always be an array, could you fix the code to clarify the intent and remove the case where `result.value` is being returned?  Let me know if you have questions regarding the fix request. Thanks! (Oh and do the same in emulation/models/orwch-loadsiz.js please!)

PR #95, opened 2017-11-17
VAM-120: ORWTPO CSARNGD

creindle:
- [X] Created all files
- [X] Passed comparison tests
- [X] Passed model tests


tests/comparison/rpcComparisonConfig.js
```diff
@@ -10,7 +10,7 @@ module.exports = {
     },
     captures: {
         filter: '^capture-.*\\.txt$',
-        path: '../metaVDP/definitions/RPC/captures',
+        path: './tests/captures',
```
mfuroyama: 🥇 Thanks for catching/fixing this!

creindle: Thank you!!

PR #97, opened 2017-11-17
VAM-113: ORWPCE1 NONCOUNT

creindle:
-[X] Created all files
-[X] Passed comparison tests
-[X] Passed model tests


services/pceMetaService/checkIfNoncountClinic.js
```diff
@@ -0,0 +1,40 @@
...
+    this.find('HospitalLocation', {
+        query: {
+            inactiveFlag: {
+                $exists: false,
+            },
+            localStationNumber: {
+                $eq: context.stationId,
+            },
+            id: {
+                $eq: context.locationId,
```
mfuroyama: Hmm...this query **should** be looking for the location associated with the **input parameter value** passed in, and not the `locationId` in the context object.

Double check the original VDP implementation and let's fix this to reflect the true intent of the RPC

services/pceMetaService/checkIfNoncountClinic.js
```diff
@@ -0,0 +1,40 @@
...
+            },
```
mfuroyama: Consider tightening up the `query` object as follows:
```Javascript
query: {
    inactiveFlag: { $exists: false },
    localStationNumber: context.stationId,
    id: context.locationId
}
```
Note that you don't need the `$eq` criteria...you can just assign the field names directly to the required match values.

creindle: Ah yeah that's a very good point!

services/pceMetaService/checkIfNoncountClinic.js
creindle: Oooooh, ok I'll check this out again. I apologize about that. And that makes sense if we added the argument to the `capture-VAM.txt` file! I know how to fix this :)

emulation/models/orwpce1-noncount.js
```diff
@@ -0,0 +1,42 @@
...
+            id: 'ORLOC',
+            parameterType: 'LITERAL',
+            required: false,
+            description: 'IEN of location',
+            example: '1',
+        }],
+
+        invokeRPC: (rpcArgs, session, callback) => {
+            if (_.isUndefined(rpcArgs) || !_.isArray(rpcArgs) || _.isEmpty(rpcArgs)) return new Error('rpcArgs is invalid: Undefined, not array, or empty');
+            if (_.isString(rpcArgs[0])) return '';
```
mfuroyama: I'm guessing you probably copied this over from the original VDP implementation. 😄  This is cool, but remember that the `invokeRPC` functions here are **asynchronous**: You need to handle errors here by calling the `callback` function with the first parameter set to the error object.

Also the second case here should also call the `callback` function with the first (error) parameter set to `null` and the second parameter set to empty string.

Let me know if you have questions

creindle: Line 32 I did copy over! :O :O So, yes, I apologize let me add those callbacks!

emulation/models/orwpce1-noncount.js
```diff
@@ -0,0 +1,42 @@
...
+        inputParameters: [{
+            id: 'ORLOC',
+            parameterType: 'LITERAL',
+            required: false,
+            description: 'IEN of location',
+            example: '1',
+        }],
+
+        invokeRPC: (rpcArgs, session, callback) => {
+            if (_.isUndefined(rpcArgs) || !_.isArray(rpcArgs) || _.isEmpty(rpcArgs)) return callback(new Error('rpcArgs is invalid: Undefined, not array, or empty'));
```
mfuroyama: One minor readability thing here.  Most times it's good to have stuff on one line for conciseness, but when the line gets **too long**, like past 100 or 120 or so characters, it's sometimes better to break the code up into multiple lines. We like to limit the max characters-per-line because:
> it's lines that have so much going on that it takes cognitive overhead to read them

In fact, there should be an ESLint rule that should have kicked and warned you about lines longer than 120 characters.

Let's break this line up in to something more manageable... 😃

creindle: I'll break this up in a moment, my vagrant box is almosttt ready!

PR #98, opened 2017-11-17
VAM-87 get Education Topics

sidjtd:
Another one!!!


emulation/models/orwordg-ien.js
```diff
@@ -0,0 +1,45 @@
...
+                example: 'ALL SERVICES',
+            },
+        ],
+        invokeRPC: (rpcArgs, session, callback) => {
+            const id = rpcArgs[0];
+            if (_.isEmpty(session.userId) || id === '' || _.isNumber(id)) return callback(null, '');
+            orderMetaService.getEducationTopics(id, session, callback);
+        },
+        formatResult: (res) => {
+            if (_.isUndefined(res) || !_.isObject(res)) return new Error('res is either undefined or not an object');
```
mfuroyama: From the way the service function is coded up, it looks like `res` should be an array? But here, it looks like `formatResult` is expecting an object.

Let me know if this isn't the case...

PR #101, opened 2017-11-20
change parameter handling to accept valid strings

kekuso:
I console.log the outputs from formatResults, which are correctly returning 'C' or 'Z', but its still not passing the Jenkins tests. Might be an issue with the Jenkins server db or tests


mfuroyama:
I'm gonna close this out, since the changes here are superceded by the other pull request

PR #102, opened 2017-11-21
Fix ORWDRA32 LOCTYPE and ORWORDG IEN RPC implementations

mfuroyama:
These changes fix the spec failures in the Jasmine comparison tests:

**ORWDRA32 LOCTYPE**
* Change the output format of the RPC results to match the expected output format

**ORWORDG IEN**
* Fix the result formatter
* Tighten up the guard conditionals in `invokeRPC` and `formatResults`
* Refactor the object map scoping


mfuroyama:
@sidjtd Thanks for reviewing this. Yeah, my brain really hurts right now, but I think I see what you're saying. I'll review the code with your comments in mind and see if I over-simplified things. Thanks!

mfuroyama:
Also, I think the changes here overlap the changes in pull request #101, I might have to do some creative merging here...

sidjtd:
Sorry maybe I can elaborate and be more succinct.  This is what I think I'm observing:

--- Input args and their respective Outputs in Mumps code :
1. * Empty string > output: -1
2. * NoArg > output: -1

3. * True Number (#) corresponding w/ valid data >  output: letter
4. * String # corresponding w/ valid data >  output: letter
5. * True # w/ NO CORRESPONDING data >  output: '' (empty str)
6. * String # w/ NO CORRESPONDING data >  output: '' (empty str)

7. * Strings other than numbers or empty string (ie: 'Hello', '14_50.50-A', '_yo_') > output > '' (empty str)

Right now I feel like the current code implementation won't meet 7's expected behavior because it'll return a -1 instead of an empty string?

mfuroyama:
@sidjtd @kekuso Okay, just pushed a fix for the edge cases. I think this should do it

PR #104, opened 2017-11-21
Vam 30 return order id via name

sidjtd:
No model spec for this one, I did final output formatting right before the service Function's callback, since all it is is returning is the tail end of the urn (the id/ien) as an integer.  So figured it doesn't need a formatResult.  LMK if you think otherwise!!!


mfuroyama:
Approved, but see the comment I left on pull request #104

PR #105, opened 2017-11-22
Vam 65 return display id via name

sidjtd:


mfuroyama:
Approved, but somehow, I feel like this RPC implementations really just wrap one of those patterns (i.e. "label-to-IEN") that we could turn into a utility. Let's revisit later, perhaps after build 1

sidjtd:
Agreed, its really silly to have separate service functions for these long term.... so I think once we rack in those implementation numbers, it only makes sense to merge these into more consolidated functions that get a collection name and string argument passed into it that will output the same result...

sidjtd:
In fact, let me try making that super fast too. Kind of a perfect day to do that today, actually.

PR #107, opened 2017-11-22
VAM-266: Create In-Depth Parameter Service Jasmine Test Suite

creindle:


mfuroyama:
Holy cannoli! 🤣  That's a lot of code!

Okay, here's a suggestion to DRY-refactor some of the code. Consider the pattern where we define a static array of result test-case objects, and then we do a `forEach` off the static array, and then embed the repeating test code in the iteration handler. So, for example, something like this:

```Javascript
    // These are the test case objects, all the test variables are self-contained in these objects
    const PARAMETER_TEST_CASES = [{
        name: 'ORB SORT METHOD',
        contextObj: context,
        entity: 'ALL',
        results: [
            { value: 'D', entity: { id: context.userId, label: userTwentyZEntity } },
            { value: 'U', entity: { id: context.domainId, label: systemEntity } },
            { value: 'D', entity: { id: packageEntityId, label: packageEntity } },
        ],
    }, {
        name: 'ORCH CONTEXT CONSULTS',
        contextObj: context,
        entity: 'ALL',
        results: [
            { value: ';;;;;0', entity: { id: context.userId, label: 'USER,TWENTY Z' } },
            { value: ';;;', entity: { id: packageEntityId, label: packageEntity } },
        ],
    }];

    // Then, iterate through the array of test cases...
    PARAMETER_TEST_CASES.forEach((testCase) => {
        const description = `Expect ParameterService to retrieve "${testCase.name}" correctly for "${testCase.entity} entity type"`;

        // ...each test case object gets it's own 'it' spec function...
        it(description, (done) => {
            // ...then we use the test case values in the ParameterService query itself...
            ParameterService.withContext(testCase.contextObj).get(testCase.name, { entity: testCase.entity })
                .then((results) => {
                    // ...but here's the biggie: create a generic expect analysis block that handles our cases.
                    // Something like this should work:
                    expect(results).toBeDefined();
                    expect(results.length).toEqual(testCase.results.length);
                    results.forEach((res, index) => {
                        expect(res.value).toEqual(testCase.results[index].value);
                        expect(res.entity.id).toEqual(testCase.results[index].entity.id);
                        expect(res.entity.value).toEqual(testCase.results[index].entity.value);
                    });
                    done();
                })
                .catch((err) => {
                    // ...and of course, manually fail all errors
                    fail(err);
                    done();
                });
        });
    });
```

What do you think? Doable? We can discuss if you like...

PR #110, opened 2017-11-27
Vam 140 tiu template access level

kekuso:
create lookupTiuTemplateAccessLevel service


emulation/models/tiu-template-access-level.js
```diff
@@ -0,0 +1,65 @@
...
+        name: 'TIU TEMPLATE ACCESS LEVEL',
+        function: 'TACCESS^TIUSRVT2',
+        parameters: [
+            'TIU PERSONAL TEMPLATE ACCESS',
+            'TIU TEMPLATE ACCESS BY CLASS',
+        ],
+        files: ['8927', '8930'],
+        domain: 'DOCUMENT',
+        labels: ['META'],
+        services: [],
```
mfuroyama: Don't forget to drop the name of the service used here!

emulation/models/tiu-template-access-level.js
```diff
@@ -0,0 +1,65 @@
...
+                required: false,
+                description: 'the user\'s location',
+                example: '1',
+            },
+        ],
+        invokeRPC: (rpcArgs, session, callback) => {
+            const template = rpcArgs[0] || '';
+            const user = rpcArgs[1] || '';
+            const location = rpcArgs[2] || '';
+            documentMetaService.lookupTiuTemplateAccessLevel(template, user, location, session, callback);
```
mfuroyama: This might a good place to consider observing our "VAM Commandments", namely: _thou shalt not pass IENs into a service function as argument values._

Consider re-working this to pull the URN conversion out of the service function and doing it here in the `invokeRPC` function instead.

PR #112, opened 2017-11-29
Refactor for vics service guidelines compliance

kekuso:
- removed any instances of 'IEN' within service implementations, changed to 'ID' or 'URN'
- moved URN extraction/creation into model emulations
- change all 'get____' services to 'lookup_____'
- add a guard/conversion to string in rpcFormatter.js


emulation/models/orwdal32-def.js
```diff
@@ -4,7 +4,7 @@
 
 const _ = require('lodash');
 const allergyMetaService = require('../../services/allergyMetaService');
-const serviceUtils = require('../../services/serviceUtils');
+const modelUtils = require('./modelUtils');
```
mfuroyama: Ahh...thanks for catching these, btw! 👍

emulation/models/orwpce-get-exam-type.js
```diff
@@ -22,13 +22,13 @@ const RPC_EMULATOR_MODEL = [
...
         invokeRPC: (args, session, callback) => {
-            pceMetaService.getExamType(session, callback);
+            pceMetaService.lookupExamType(session, callback);
         },
         // Format pureJS and mumps version
         formatResult: (res) => {
             if (_.isUndefined(res) || !_.isArray(res)) {
                 return new Error('res is either undefined or not array');
-            } return res.map(element => `${element.ien}^${element.label}`);
+            } return res.map(element => `${element.id}^${element.label}`);
```
mfuroyama: Hmm...I could be wrong here, but I thinking we might want to do more than just rename the object field names from `ien` to `id`. If we honor the "letter of the commandments" here, I think we might have to go deeper and perform actual conversions of the object field *values* from IENs to URNs.

I'm totally up for discussing this, though, if you have thoughts about alternative implementation ideas.

services/orderMetaService/checkIfIsClinicalLocation.js
```diff
@@ -4,28 +4,27 @@ const serviceUtils = require('../serviceUtils');
 const _ = require('lodash');
 
 /**
- * @summary Checks if incoming argument is a valid IEN for a clinical location
- * @param  {Number}   locationIEN  IEN for file 44 used to search for a Hospital location
+ * @summary Checks if incoming argument is a valid ID for a clinical location
+ * @param  {Number}   locationIEN  ID for file 44 used to search for a Hospital location
```
mfuroyama: This is manini, but we can probably change `locationIEN` here to `locationURN` to match the function signature

services/pceMetaService/askPCE.js
```diff
@@ -16,17 +16,14 @@ const ParameterService = require('../parameterService');
...
-function askPCE(userId, location, context, callback) {
+function askPCE(userURN, locationURN, context, callback) {
     // If incoming argument is a valid number, use that as Id arg. Else, use current users Id
-    const validNumber = (_.isNumber(parseInt(userId, 10)));
-    const contextId = serviceUtils.getIenFromId(context.userId);
-    const idToCheck = (!validNumber) ? contextId : userId;
-
-    const urnString = serviceUtils.createURNFromParams(200, idToCheck, context.stationId);
-    const stationId = context.locationStationId;
+    const validNumber = (_.isNumber(parseInt(userURN, 10)));
```
mfuroyama: If we're truly expecting a URN-type user identifier now, we'll have to perform a `serviceUtils.getIdFromUrn` before we do the `parseInt`?

services/pceMetaService/checkIfNoncountClinic.js
```diff
@@ -5,24 +5,22 @@ const _ = require('lodash');
 
 /**
  * @summary Is a given HOSPITAL LOCATION (file 44) a non-count clinic?  (DBIA #964)
- * @param  {Number}   locationIEN  Hospital IEN
+ * @param  {Number}   locationIEN  Hospital ID
```
mfuroyama: This one too...`locationIEN` --> `locationURN`?

services/tests/checkIfNoncountClinic-spec.js
```diff
@@ -9,17 +9,16 @@ describe('checkIfNoncountClinic asynchronous specs', () => {
...
+        modelFunction.checkIfNoncountClinic('urn:vista:6100:44:1', context, (err, res) => {
+            if (err) console.log(err);
             expect(res).toEqual(jasmine.any(Boolean));
             done();
         });
     });
     it('Passing in non number string returns empty string', (done) => {
         modelFunction.checkIfNoncountClinic('foo', context, (err, res) => {
-            if (err) fail();
+            if (err) console.log(err);
```
mfuroyama: Actually, can we revert this back to `fail`?  This and all other *unexpected* error handler cases in our specs should manually `fail` out of the test case instead of just printing to console.

As we do code reviews, let's keep an eye out for this...

tests/captures/capture-VAM.txt
```diff
@@ -48,9 +48,5 @@
 {
   "name": "ORWPCE1 NONCOUNT",
   "args": [1]
-},
-{
-  "name": "ORIMO ISCLOC",
-  "args": [1]
```
mfuroyama: Just out of curiosity, what happened to this poor RPC sample getting deleted out?

kekuso: Ah, I think there's already an RPC sample in another capture file, but I can add it back in if the intent was to test that specific param (1)

services/tests/checkIfNoncountClinic-spec.js
kekuso: OK, will do that in all future implementations too!

emulation/models/orwpce-get-exam-type.js
kekuso: Wouldn't that break our tests where we compare the outputs of formatResults to the rpcRunner (native) results?

mfuroyama: Yeah, it would break the test results as-is. But (and we don't need to do it this go-around) the ideal fix here would be to modify **both** the service function (to return URNs in the `id` field) **and** the `formatResult` function (to parse the IENs out of the results)

I think this is a good first pass though, but eventually, we'll have to further purify the service and model functionality

tests/captures/capture-VAM.txt
mfuroyama: Nah, no need add back...I was just curious...thanks!

emulation/models/orwpce-get-exam-type.js
kekuso: Ahh I see, and I agree!

PR #113, opened 2017-11-30
Vam 46 orwcom ptobj

kekuso:
- create lookupPatientCOMObject service and test specs
- create orwcom-ptobj model and test specs
- minor model fixes


emulation/models/orwcom-ptobj.js
```diff
@@ -0,0 +1,41 @@
...
+        labels: ['META'],
+        services: ['orderMetaService'],
+        inputParameters: [],
+        invokeRPC: (args, session, callback) => {
+            orderMetaService.lookupPatientCOMObject(session, callback);
+        },
+
+        formatResult: (res) => {
+            if (!res || _.isEmpty(res)) return new Error('expected Object param in formatResult');
+            if (res.error) return 0;
```
mfuroyama: Hmm, I'm wondering if we should return a wrapped error instead of plain ol' `0` if there's an error?

emulation/models/orwcom-ptobj.js
```diff
@@ -0,0 +1,41 @@
...
+            const resArr = Object.keys(res).map((key) => {
+                if (key === 'id') return modelUtils.getIenFromId(res[key]);
```
mfuroyama: Consider a ternary here...

services/orderMetaService/lookupPatientCOMObject.js
```diff
@@ -0,0 +1,50 @@
...
+ * @instance
+ */
+function lookupPatientCOMObject(context, callback) {
+    const service = this;
+    ParameterService.get('ORWCOM PATIENT SELECTED', { entity: ['USR', 'SRV', 'DIV', 'SYS'] })
+        .then((res) => {
+            // check the Parameter ORWCOM PAITENT SELECTED for an existing COM Object id
+            const paramVal = _.get(res[0], 'paramVal', '');
+            // if it does not exist, return error Object
+            if (!paramVal || !parseInt(paramVal, 10)) return callback(null, { error: 'COM Parameter not found' });
```
mfuroyama: So, these `return` statements might not have the effect you're expecting here.  If you don't return a promise from within the `resolve` handler function, the result returned gets wrapped in another Promise and passed to the next `then` handler.

I'm pretty sure you can see this in action if you add a `console.log` statement to the second `then` handler, then hard-code `paramVal` to an invalid value. The `console.log` will **always** be triggered.

There are several possibilities I can think of to consider to handle this specific promise chaining case:
1. In the first `then` handler, instead of calling `callback`, return the error message, and in the second `then` handler, check `res` for error strings, and call `callback` there.
2. Use the callback option in `ParameterService.get` instead of promise chaining.

We can talk about this tomorrow, if this doesn't make sense. It's kinda tricky, but once you resolve this, it'll be a good example for the rest of the team.

services/tests/lookupPatientCOMObject-spec.js
```diff
@@ -0,0 +1,22 @@
...
+    beforeAll(() => {
+        const res = serviceTestUtils.beforeAllScript();
+        context = res.context;
+    });
+
+    it('Empty arg should return COM Object', (done) => {
+        orderMetaService.lookupPatientCOMObject(context, (err, res) => {
+            if (err) fail();
+            expect(res).toBeDefined();
+            expect(res).toEqual(jasmine.any(Object));
```
mfuroyama: Might be a good idea to check for specific fields here (e.g. `id`, etc) instead of just whiffing with the equivalent of an `isObject` test...sorry, I gotta be more pedantic, remember? 😆

emulation/models/orwcom-ptobj.js
kekuso: Is there a way to do this without breaking the comparison-tests? The native result is 0.

mfuroyama: Oh! If that's the case then yeah, this should be good! No worries...

PR #115, opened 2017-12-01
VAM-96 ORWPCE GET EDUCATION TOPICS

pcchanghawaii:
please let me know if this is not what you want for feature branch/pull request. Do I need to put reviewers and assignees to you/me each time?

Thanks!


creindle:
Hi Pei!

services/tests/getPceEducationTopics-spec.js
```diff
@@ -0,0 +1,50 @@
...
+            expect(res[0].id).toBeDefined();
+            expect(res[0].name).toBeDefined();
+        }
+
+
+    };
+
+    it('Empty arg should return array of Topics objects', (done) => {
+        pceMetaService.getEducationTopics((err, res) => {
+            if (err) console.log(err);
```
kekuso: We should probably call `fail()` here instead of just `console.log()`
This and all other unexpected error handler cases in our specs should manually fail out of the test case instead of just printing to console.
Some of our test specs need to be revised to reflect these changes too since we only recently started using `fail()`

services/tests/getPceEducationTopics-spec.js
```diff
@@ -0,0 +1,50 @@
...
+    });
+
+
+
+    const testReturnsTopicsArray = (res) => {
+        expect(res).toBeDefined();
+        expect(res).toEqual(jasmine.any(Array));
+
+        if (res[0]) {
+            let startsWithA = false;
```
kekuso: hmm I think a more generic test would be better here, in the off-chance that there are no Education Topics that begin with the letter A...
I would do something like `resCopy.sort() === res` to check for an alphabetized array

services/pceMetaService/getEducationTopics.js
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+function getEducationTopics(callback) {
+    /* Get all education topics */
+    this.find('EducationTopics', {
+        query: { inactiveFlag: { $exists: false } },
+        projection: serviceUtils.createProjection(['id', 'label']),
+    })
+        .then((topicArr) => {
+            const res = _(topicArr)
```
kekuso: +1 for lodash chaining!

services/pceMetaService/getEducationTopics.js
```diff
@@ -0,0 +1,36 @@
+'use strict';
```
creindle: This was a recent change but the title of the file should be changed from **get** to something like **lookup**. Here is a reference to the recent change: https://github.com/vistadataproject/VICSServer/wiki/Guidelines-for-VICS-Services

services/tests/getPceEducationTopics-spec.js
```diff
@@ -0,0 +1,50 @@
+'use strict';
```
creindle: I believe this spec file is for the `getEducationTopics.js` correct? Therefore the spec file should match it by simply adding the `-spec` near the end of the name. The new name for this service spec file would become `getEducationTopics-spec.js`

emulation/models/orwpce-get-education-topics.js
```diff
@@ -0,0 +1,32 @@
...
+        returnType: 'ARRAY',
+        files: ['9999999.09'],
+        parameters: [],
+        labels: ['META'],
+        services: ['pceMetaService'],
+        inputParameters: [],
+        invokeRPC: (args, session, callback) => {
+            pceMetaService.getEducationTopics(callback);
+        },
+        formatResult: (res) => {
```
creindle: Since this model file contains the method `formatResults` and has logic in it, we must have a spec file to test the contents of the results within. So, the name of this file would be `orwpce-get-education-topics-spec.js` and would be located in `VICSServer/emulation/models/tests`.

emulation/models/orwpce-get-education-topics.js
```diff
@@ -0,0 +1,32 @@
...
+            if (res == null || !Array.isArray(res)) return new Error('formatResult only accepts arg of type Array');
```
creindle: Good job on adding the check! I would always forget this!!

services/pceMetaService/getEducationTopics.js
creindle: I agree! +1

creindle:
Also, in the title of this pull request, we include "VAM-#" so that it would trigger in Jira that this issue is "In Progress".

services/tests/getPceEducationTopics-spec.js
pcchanghawaii: there's already the lookupEducationTopics-spec.js file, which is for orderMetaService.  That one looks for for a particular education topic while the getEducationTopics pulls a list of all education topics.  Maybe I will put Pce on both files, i.e.,  lookupPceEducationTopics.js & lookupPceEducationTopics-spec.js      Thanks for pointing it out!

services/tests/getPceEducationTopics-spec.js
pcchanghawaii: good point. I copied it from the other spec in the pceMetaService. In fact, all other specs in this folder use console.log. I will modified them as well.

services/pceMetaService/getEducationTopics.js
pcchanghawaii: Thanks! I think I copied the code snippet from the other file(s) from the pceMetaService folder as well. I will update the other files that require the implicit chaining.

pcchanghawaii:
Hi there,  I've updated many files based on the the above comments.  would you or the team please review again?  Do I need to submit another pull request?

Please let me know.
Thanks!!  

Pei

creindle:
You do not need to make another pull request, we are able to approve the changes on here. Although you can merge your branch into the master branch at anytime, it is best to wait until you have the approval from every reviewer you requested it from. Then you'll see the "Changes approved" and that means you're good to go!

PR #117, opened 2017-12-01
Parameter Service In-Depth Jasmine Tests Fixes

mfuroyama:
This code review is pretty much just FYI, @creindle. Take a look, for your own piece of mind, then approve and I'll merge in.

The test failures pretty much came down to two issues:
1. There was actually a mismatch in the way labels were stored between the `Parameter` and the `ParameterDefinition` collections. in the `Parameters` collection, labels were stored as `ORK SYSTEM ENABLE_DISABLE` where they're stored as `ORK SYSTEM ENABLE/DISABLE` in the `ParameterDefinition` collection.
2. The location-type test cases for `ORWPCE ASK ENCOUNTER UPDATE` were slightly off. Changing the test cases to match the existing data records fixed the failures.


services/parameterService.js
```diff
@@ -86,7 +86,7 @@ class ParameterService extends BaseService {
             }
 
             // Make sure we're only pulling parameters for the station specified in the context
-            const query = { label: name };
+            const query = { label: name.split('_').join('/') };
```
creindle: I thought you fixed this already?? :O Ok, I should've known!

creindle: I distinctly remember you talking about how you fixed it thats why.

services/tests/parameterServiceInDepthALL-spec.js
```diff
@@ -43,6 +43,15 @@ describe('Parameter Services Jasmine "ALL" tests', () => {
         stationId: '9999',
         locationId: 'urn:vista:9999:44:123',
     };
+    const contextICUS = {
```
creindle: What is ICUS?

services/tests/parameterServiceInDepthSingle-spec.js
```diff
@@ -1240,14 +1248,14 @@ describe('Parameter Services Jasmine "Single Entity" tests', () => {
     * Entity Length 1
     * Value (type)  '5, 6, 7' (set of codes (0:Primry/Data Needed; 1:Primry/Outpt; 2:Primry Always; 3:Data Needed; 4:Outpt; 5:Always; 6:Never; 7:Disable))
     */
-    xit('Expect ParameterService to retrieve "ORWPCE ASK ENCOUNTER UPDATE" correctly for PACKAGE Entity Type', (done) => {
-        ParameterService.withContext(context).get('ORWPCE ASK ENCOUNTER UPDATE', { entity: 'LOC' })
+    it('Expect ParameterService to retrieve "ORWPCE ASK ENCOUNTER UPDATE" correctly for PACKAGE Entity Type', (done) => {
+        ParameterService.withContext(contextICUS).get('ORWPCE ASK ENCOUNTER UPDATE', { entity: 'LOC' })
```
creindle: Ooooh, should I have looked closer into the MongoDB for a correct user.id? Or is this something new?

services/parameterService.js
mfuroyama: Hmm, yeah, there's a couple places where the comparison is made.

Eventually, we might actually go through and normalize the data sets, but bottom line is that it's good we caught this here!  This is exactly the reason why we wanted to build these tests!

services/tests/parameterServiceInDepthALL-spec.js
mfuroyama: `ICUS` is short for `ICU-S`, which is the label value of the location (44) entity that contains a valid value. The original `context` object location entity doesn't contain values for the named parameter

services/tests/parameterServiceInDepthSingle-spec.js
mfuroyama: See the above comment. I created a new session context object for this particular test to get the test specs to return values.

services/parameterService.js
creindle: Awesome :)

PR #119, opened 2017-12-02
VAM-276: Test reduction of PKG where no map exists for a parameter

creindle:
- Tested with parameters `GMV DLL VERSION` and `GMV GUI VERSION`.
- Tests do not pass at the moment.


mfuroyama:
Hey @creindle! So check out the changes I pushed to the branch. They contain:
* One minor change to the ParameterService code. This prevents a entity search on 'PKG' for parameter types that don't belong to a package from returning values.
* Test other failing tests by modifying the specs to get the expected values.

Check out the changes, and if you're okay with them and they make sense, let's move them in.

PR #121, opened 2017-12-04
Vam 129 orwu validsig

kekuso:
- created validateESignature service and tests
- created orwu-validsig model
- added Cipher to serviceUtils


services/cprsMetaService/validateESignature.js
```diff
@@ -0,0 +1,52 @@
+'use strict';
+
+const serviceUtils = require('../serviceUtils');
+const _ = require('lodash');
+
+/**
+ * @summary Validates a broker encrypted electronic signature
+ * @param  {[type]}   encryptedString The encrypted string to decrypt and compare to current user's electronic signature
```
mfuroyama: Shoot, sorry, I have to put on my pedantic hat here, but can we change `[type]` here to `String`? Otherwise, it'll break the doc generator...

PR #122, opened 2017-12-04
VAM-83: ORWOR PKISITE

creindle:
[X] Made correct files
[X] Passes comparison tests


services/orderMetaService/isEncryptionEnabled.js
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * @summary determines if PKI is turned on at the site by verifying electronic Prescribing of Controlled Substances (ePCS) parameters used by CPRS
+ * @param  {Object}   context  Session context information associated with the service call.
+ * @param  {Function}   callback  Function called on process completion, with standard (err, result) signature.
+ * @return {Number}   isEpcsEnabled  enableEpcs field 1 (true) or 0 (false)
+ *
+ * @memberof OrderMetaService
+ * @instance
+ */
+function isEncryptionEnabled(context, callback) {
```
mfuroyama: I'm thinking we can probably remove the `context` parameter out of the function signature here, since we don't use it in the function...

creindle: :O Really?? Ok!

mfuroyama: Yeah, in our service functions, the signatures should only include stuff necessary to run the function. We can leave out stuff we don't use.

services/orderMetaService/isEncryptionEnabled.js
```diff
@@ -0,0 +1,29 @@
+'use strict';
+
+const serviceUtils = require('../serviceUtils');
+const _ = require('lodash');
+
+/**
+ * @summary determines if PKI is turned on at the site by verifying electronic Prescribing of Controlled Substances (ePCS) parameters used by CPRS
+ * @param  {Object}   context  Session context information associated with the service call.
```
mfuroyama: Don't forget to clear out the parameter in the JSDoc comment too!

PR #123, opened 2017-12-04
VAM-281 Correct filename

sidjtd:
I noticed that test-comparisons was still failing regarding a getEducationTopics name issue but... I think the team members said that is expected behavior because of Pei related something? Pushing this through so I can get on moving!


sidjtd:
Haha, if the under-carriage of the bus needs maintenance, that's usually the right place to be tossed into!

PR #125, opened 2017-12-04
VAM-146: XWB GET BROKER INFO

creindle:
[X] Passes comparison tests
[X] Created all necessary files


mfuroyama:
Hey! If you have a moment, let's talk about this. I think there's an opportunity to reuse the `lookupKernelSystemParameters` service function in the "SystemMeta" service...

emulation/models/tests/xwb-get-broker-info-spec.js
```diff
@@ -0,0 +1,40 @@
...
+        const res = unitTestUtils.beforeAllScript(rpcEmulatorModel);
+        rpc = res;
+        context = res.context;
+    });
+
+    it('invokeRPC: Should return response object containing brokerActivityTimeout', (done) => {
+        rpc.run([], context, (err, res) => {
+            if (err) fail(err);
+            expect(res).toBeDefined();
+            expect(_.has(res, 'brokerActivityTimeout')).toEqual(true);
```
mfuroyama: This is minor, but let's switch this `toEqual` out here to `toBeDefined`. It's subtle, but it's more explicit in the resulting reports

services/tests/lookupDisplayGroup-spec.js
```diff
@@ -1,7 +1,7 @@
 'use strict';
 
 const _ = require('lodash');
-const modelFunction = require('../orderMetaService');
+const modelFunction = require('../pceMetaService');
 const serviceTestUtils = require('./serviceTestUtils');
 
 describe('lookupDisplayGroup asynchronous specs', () => {
```
mfuroyama: We can delete this file out of the feature branch, since it's not used anymore

PR #126, opened 2017-12-05
VAM-135: TIU GET PRINT NAME

creindle:
[X] Passes comparison tests
[X] Created necessary files


emulation/models/tiu-get-print-name.js
```diff
@@ -0,0 +1,37 @@
...
+const _ = require('lodash');
+
+const RPC_EMULATOR_MODEL = [
+    {
+        id: 'TIU_GET_PRINT_NAME',
+        name: 'TIU GET PRINT NAME',
+        function: 'GETPNAME^TIUSRVP1',
+        domain: 'DOCUMENT',
+        parameters: [],
+        files: ['8925.1'],
```
mfuroyama: Looks like you're missing the `services` array field. If anything, this should have failed the "model validity" test... 😉

services/documentMetaService/lookupDocumentPrintName.js
```diff
@@ -0,0 +1,34 @@
...
+ * @param  {Number}   documentNumber  Document to search for
+ * @param  {Object}   context  Session context information associated with the service call.
+ * @param  {Function}   callback Function called on process completion, with standard (err, result) signature.
+ * @return {String}   getPrintName  printName based on document if exists or label
+ *
+ * @memberof DocumentMetaService
+ * @instance
+ */
+function lookupDocumentPrintName(documentNumber, context, callback) {
+    const documentId = serviceUtils.createURNFromParams('8925_1', documentNumber, context.stationId);
```
mfuroyama: Let's do the IEN-to-URN conversion in the model RPC `invokeRPC` function.  Remember that, according to the [VAM Commandments](https://github.com/vistadataproject/VICSServer/wiki/Guidelines-for-VICS-Services), service functions should always deal with URNs...

services/documentMetaService/lookupDocumentPrintName.js
```diff
@@ -0,0 +1,34 @@
...
+    this.find('TiuDocumentDefinition', {
+        query: { id: documentId },
+        projection: serviceUtils.createProjection(['id', 'printName']),
+    }).then((res) => {
+        let getPrintName = _.get(res[0], 'printName', '');
+        if (getPrintName) return callback(null, getPrintName);
+        getPrintName = _.get(res[0], 'label', 'UNKNOWN');
```
mfuroyama: If you want this to work, you'll probably have to include `'label'` in the query projection listing above.

services/documentMetaService/lookupDocumentPrintName.js
```diff
@@ -0,0 +1,34 @@
...
+        getPrintName = getPrintName === 'UNKNOWN' ? 'UNKNOWN' : _.startCase(_.toLower(getPrintName));
```
mfuroyama: Hmm...I think there's room to tighten up the previous two lines. Consider the following replacement:
```Javascript
const label = _.get(res[0], 'label');
const result = label ? _.startCase(_.toLower(label)) : 'UNKNOWN';
```

I feel like this way, the intent of to code is communicated a little bit more clearly. My suggestion makes it clear that the `'UNKNOWN'` result is your fall-back value, and not something you're checking for. Hopefully, that made sense...

emulation/models/tiu-get-print-name.js
```diff
@@ -0,0 +1,37 @@
...
+            id: 'TIUTYPE',
+            parameterType: 'LITERAL',
+            required: true,
+            description: 'Pointer to TIU DOCUMENT DEFINITION FILE (#8925.1)',
+            example: '1',
+        }],
+
+        invokeRPC: (rpcArgs, session, callback) => {
+            if (_.isEmpty(rpcArgs)) throw new Error('Empty arguments');
+            if (parseInt(rpcArgs[0], 10) <= 0) return 'UNKNOWN';
```
mfuroyama: See my comment in the service function below. Let's do the IEN-to-URN conversion here

emulation/models/tiu-get-print-name.js
creindle: 😭

services/documentMetaService/lookupDocumentPrintName.js
creindle: That makes a lot more sense now!

services/documentMetaService/lookupDocumentPrintName.js
creindle: SO weird I thought I had that field there! :O :O

services/documentMetaService/lookupDocumentPrintName.js
creindle: This is perfect! I was really trying to figure out how to tighten the code without compromising the integrity of the nonClinicialRPC implementation!

services/documentMetaService/lookupDocumentPrintName.js
```diff
@@ -0,0 +1,33 @@
...
+ * @summary Returns a string containing the Print Name of the corresponding Document Definition.
+ * @param  {Number}   documentNumber  Document to search for
+ * @param  {Object}   context  Session context information associated with the service call.
+ * @param  {Function}   callback Function called on process completion, with standard (err, result) signature.
+ * @return {String}   getPrintName  printName based on document if exists or label
+ *
+ * @memberof DocumentMetaService
+ * @instance
+ */
+function lookupDocumentPrintName(documentNumber, context, callback) {
```
mfuroyama: Ahh, sorry to be so pedantic about this, but let's change the `documentNumber` parameter here to `documentId`.  We'll also want to change `documentNumber` in the JSDoc comments, as well as the argument type (from Number to String)

creindle: Don't be sorry! No worries! I need to understand the specific terms!

PR #127, opened 2017-12-06
Vam 69 orwdx wrlst

kekuso:
- created lookupWritingOrdersList service and tests
- created orwdx-wrlst model and tests
- fix instances of ParameterService without `withContext()` call


scripts/boilerTemplates.js
```diff
@@ -19,7 +19,7 @@ describe('${custom} asynchronous specs', () => {
 
     it('Empty arg returns .....', (done) => {
         modelFunction.${custom}([], context, (err, res) => {
-            if (err) console.log(err);
+            if (err) fail();
```
mfuroyama: :+1:

emulation/models/orwdx-wrlst.js
```diff
@@ -0,0 +1,76 @@
...
+const RPC_EMULATOR_MODEL = [
+    {
+        id: 'ORWDX_WRLST',
+        name: 'ORWDX WRLST',
+        function: 'WRLST^ORWDX',
+        domain: 'ORDER',
+        files: ['101.41', '200'],
+        parameters: ['ORWOR WRITE ORDERS LIST', 'ORWDX WRITE ORDERS LIST'],
+        labels: ['META', 'DISPLAY GROUP', 'RESULTS-UNSORTED'],
+        services: ['orderMetaService'],
```
mfuroyama: Hmm...I think this might be a 'cprsMetaService' RPC, tbh, since it looks like we're mostly dealing with display type stuff?

emulation/models/orwdx-wrlst.js
```diff
@@ -0,0 +1,76 @@
...
+            if (!validObj) return new Error('invalid Object properties');
+            const dialogListArr = res.map((el) => {
+                let dialogListString = [
+                    modelUtils.getIenFromId(el.id),
+                    el.windowFormId || '',
+                    modelUtils.getIenFromId(el.displayGroupId) || 0,
+                    TYPES[el.type] || '',
+                ].join(';');
+                dialogListString += `^${el.displayText || ''}`;
+                return dialogListString;
```
mfuroyama: This is minor, but you _might_ want to consider using a string literal template here to get your return value:
```Javascript
return `${dialogListString}^${el.displayText || ''}`;
```
Side benefit: if you go this route, you can make `dialogListString` const!

emulation/models/orwdx-wrlst.js
```diff
@@ -0,0 +1,76 @@
...
+        invokeRPC: (rpcArgs, session, callback) => {
+            const location = rpcArgs[0] ? modelUtils.createURNFromParams('44', rpcArgs[0], session.stationId) : session.locationId;
+            orderMetaService.lookupWritingOrdersList(location, session, callback);
+        },
+        formatResult: (res) => {
+            if (!res || !Array.isArray(res)) return new Error('formatResult param must be an Array');
+            const validObj = res.every((el) => {
+                const keys = Object.keys(el);
+                const hasKey = keys.every(key => KEYS.indexOf(key) >= 0);
+                return hasKey;
```
mfuroyama: Hmm, what about the following:
```Javascript
return Object.keys(el).every(key => _.includes(KEYS, key));
```

emulation/models/orwdx-wrlst.js
kekuso: 👍

emulation/models/orwdx-wrlst.js
kekuso: 👍 👍

PR #128, opened 2017-12-06
VAM-61: ORWDPS32 VALROUTE

creindle:
This RPC was a little bit more tough then what I was used to, can I have eagle eyes for this one and let me know if I interpreted the nonClinicialRPC version of this (f110_orwdps32Valroute.js) incorrectly? :D


services/medMetaService/lookupIENForValidRoute.js
```diff
@@ -0,0 +1,34 @@
...
+ * @param  {Function} callback Function called on process completion, with standard (err, result) signature.
+ * @return  {Object}   medicationRoute  Medication Route containing IEN and abbreviation
+ *
+ * @memberof MedMetaService
+ * @instance
+ */
+function lookupIENForValidRoute(routeName, context, callback) {
+    const upperCaseArg = routeName.toUpperCase();
+    this.find('MedicationRoutes', {
+        query: { localStationNumber: context.stationId, label: upperCaseArg, packageUse: 'ALL PACKAGES' },
```
kekuso: I think you only need a projection of the `id`, `label`, and `abbreviation`

kekuso: Also, I know the comments in the original emulation mentions 'only getting records with `packageUse`', but I think that was an artifact and was supposed to be deleted (was probably my bad!).

So you can remove the `packageUse: 'ALL PACKAGES'` query

services/tests/lookupIENForValidRoute-spec.js
```diff
@@ -0,0 +1,53 @@
...
+        modelFunction.lookupIENForValidRoute('ORAL', context, (err, res) => {
+            if (err) fail(err);
+            expect(res).toBeDefined();
+            expect(res).toEqual(jasmine.any(Object));
+            expect(res.matchingIen).toEqual('1');
+            expect(res.abbreviation).toEqual('PO');
+            done();
+        });
+    });
+    it('Valid string "SL" returns IEN 2 for route', (done) => {
```
kekuso: So here is another case where the original emulation is probably not complete...'SL' (not 'SUBLINGUAL') should be a valid param and should return the same thing as passing in 'SUBLINGUAL'. Therefore, you'll probably need to tweak your query in `lookupIENForValidRoute` so that it checks against both the `name` field and the `abbreviation` field.

emulation/models/orwdps-32-valroute.js
```diff
@@ -0,0 +1,46 @@
...
+            id: 'X',
+            parameterType: 'LITERAL',
+            required: true,
+            description: 'Valid string argument of 51.2 Medication Route returns abbreviation and IEN',
+            example: 'ORAL',
+        }],
+        invokeRPC: (rpcArgs, session, callback) => {
+            if (typeof rpcArgs[0] === 'boolean') return new Error('Boolean type argument not valid.');
+            if (!_.isString(rpcArgs[0]) || rpcArgs[0] === '') return 0;
+            medMetaService.lookupIENForValidRoute(rpcArgs[0], session, callback);
```
kekuso: You miiight want to reconsider the name of this service, I think something like `lookupMedicationId` might be more appropriate
(and would conform to the VAM Commandments :))

emulation/models/tests/orwdps-32-valroute-spec.js
```diff
@@ -0,0 +1,43 @@
+'use strict';
+
+const rpcEmulatorModel = require('../orwdps-32-valroute').rpcEmulatorModel;
+const unitTestUtils = require('./unitTestUtils');
+
+describe('ORWDPS32VALROUTE asynchronous specs', () => {
```
mfuroyama: Sorry to be pedantic here, but can we change the `describe` description text here to:
```Javascript
'ORWDPS32 VALROUTE asynchronous specs'
```

Adding the space in the RPC name makes it a little easier to identify the RPC under test in generated reports

emulation/models/tests/orwdps-32-valroute-spec.js
```diff
@@ -0,0 +1,43 @@
...
+    beforeAll(() => {
+        const res = unitTestUtils.beforeAllScript(rpcEmulatorModel);
+        rpc = res;
+    });
+
+    const testReturnsArray = (res) => {
+        expect(res).toBeDefined();
+        expect(res).toEqual(jasmine.any(String));
+        const record = res.split('^');
+        expect(record.length === 2).toEqual(true);
```
mfuroyama: Here, instead of checking for `length === 2` to be true, consider checking for `length` to be 2. Again, subtle differences, but you get much better feedback in your test results/reports if your test conditions are as explicit as possible.

services/medMetaService/lookupIENForValidRoute.js
```diff
@@ -0,0 +1,34 @@
...
+
+const serviceUtils = require('../serviceUtils');
+const _ = require('lodash');
+
+/**
+ * @summary Returns the IEN for a route if the name is valid.
+ * @param  {String}   routeName  Route name to validate
+ * @param  {Object}   context  Session context information associated with the service call.
+ * @param  {Function} callback Function called on process completion, with standard (err, result) signature.
+ * @return  {Object}   medicationRoute  Medication Route containing IEN and abbreviation
```
mfuroyama: I'll get to the actual code here in a bit, but again, remember that service functions should *not* be dealing with IENs

services/medMetaService/lookupIENForValidRoute.js
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * @memberof MedMetaService
+ * @instance
+ */
+function lookupIENForValidRoute(routeName, context, callback) {
```
mfuroyama: Consider renaming the service function to something like `lookupValidRoute`. Again, we want to eliminate references to IENs in our service functions...

services/medMetaService/lookupIENForValidRoute.js
```diff
@@ -0,0 +1,34 @@
...
+    const upperCaseArg = routeName.toUpperCase();
+    this.find('MedicationRoutes', {
+        query: { localStationNumber: context.stationId, label: upperCaseArg, packageUse: 'ALL PACKAGES' },
+    }).then((res) => {
+        const labelIen = _.get(res[0], 'id', 0);
+        const medicationRoute = {
+            matchingIen: serviceUtils.getIdFromUrn(labelIen),
```
mfuroyama: Let's return the id URN here, and defer the IEN extraction to the `formatResult` function. As noted earlier, we want to ensure that we're not doing anything IEN related in the service functions, if at all possible.

emulation/models/orwdps-32-valroute.js
mfuroyama: 👍 I think I made a similar comment below in the service function implementation!

services/medMetaService/lookupIENForValidRoute.js
creindle: Hahahahha! Thanks for the clarification!

emulation/models/tests/orwdps-32-valroute-spec.js
creindle: I completely understand! I'll be on the lookout for this for my other RPCs

emulation/models/tests/orwdps-32-valroute-spec.js
creindle: I understand! I'll try to be more careful next time!

services/tests/lookupIENForValidRoute-spec.js
creindle: 😭

PR #131, opened 2017-12-06
VAM-287: Fix Service Call that implements ORWOR PKISITE

creindle:
[X] Changed service function and file names to isEPCSEnabled
[X] Returns boolean from service now (Now I understand it as trying to take out MUMPS-y formatting from our service functions) and put it in formatResults
[X] Passed in context to service function and does mongoDB query based on sessionId


emulation/models/tests/orwor-pksite-spec.js
```diff
@@ -0,0 +1,26 @@
...
+
+    beforeAll(() => {
+        const res = unitTestUtils.beforeAllScript(rpcEmulatorModel);
+        rpc = res;
+        context = res.context;
+    });
+
+    it('formatResult: Input response of Boolean true should return MUMPS equivalent 1', () => {
+        const res = rpc.formatResult(true);
+        expect(res === 1).toEqual(true);
```
mfuroyama: Let's change this `expect` statement (and the one below) to use the following pattern:
```Javascript
expect(res).toEqual(1);
```
:wink:

creindle: Noooooo!! :O :O I should've caught that DUHHH!!

PR #133, opened 2017-12-07
Vam 49 orwcv1 coversheet list

kekuso:
- created orwcv1 coversheet list models and tests
- created lookupCoverSheetReports service and tests


emulation/models/orwcv1-coversheet-list.js
```diff
@@ -0,0 +1,106 @@
...
+    'param1',
+    'qualifier',
+    'tabPositions',
+    'pieces',
+    'detailReport',
+    'index',
+];
+
+// Qualifiers Field Range
+const QUALIFIERS = {
```
mfuroyama: I forget...were these values hard-coded in the MUMPS?

services/orderMetaService/lookupCoverSheetReports.js
```diff
@@ -0,0 +1,97 @@
+'use strict';
+
+const serviceUtils = require('../serviceUtils');
+const ParameterService = require('../parameterService');
+const _ = require('lodash');
+
+// OE/RR Report Properties
+const PROPERTIES = [
```
mfuroyama: Just a thought, but I noticed that you have two identical statically-defined constant arrays, here and in the RPC model code. Consider removing one array definition from the code and having that definition refer to the single source of truth. In general, this makes for more maintainable code.

services/orderMetaService/lookupCoverSheetReports.js
```diff
@@ -0,0 +1,97 @@
...
+                const detailObj = _.find(detailObjList, (elem) => {
+                    if (!elem.id || !_.has(el, 'detailReport.id')) return false;
+                    return elem.id === el.detailReport.id;
+                });
+                const newDetailObj = el;
+                newDetailObj.detailReport = _.get(detailObj, 'rpc.label', '');
+                return newDetailObj.detailReport ? newDetailObj : el;
+            });
+            return callback(null, finalReportList);
+        })
```
mfuroyama: This is find as-is, but let's keep this one in the back of our minds; we might want to consider looking into tweaking the 'OerrReport' collection schema by replacing the `detailReport` IDs with MongoDB references. This way, we only have to do one query, and MongoDB does the second query work for us...

emulation/models/orwcv1-coversheet-list.js
kekuso: The field ID's were hard-coded, but in the MVDM emulation we used a function called emulatorUtils.getFieldRange to get the range of field values...but that was with VDM if I recall correctly. We need access to a certain set of schemas to get the field ranges instead of hard-coding them. Let me know if you'd like me to do that and we can discuss!

kekuso: The MUMPS code: http://code.osehra.org/dox/Routine_ORWCV1_source.html

services/orderMetaService/lookupCoverSheetReports.js
kekuso: Actually they're slightly different, this one has 'id', the other doesn't.
But I can tweak the code to handle that.
The only issue is where to store the array...I'm having trouble exporting it from `lookupCoverSheetReports`, does this have to do with singletons? Or am I just importing/exporting it incorrectly?
exporting from `lookupCoverSheetReports.js`: `module.exports = { lookupCoverSheetReports, PROPERTIES };`
importing in `orwcv1-coversheet-list.js`:
`const PROPERTIES = orderMetaService.PROPERTIES`

mfuroyama: Ahh, okay, if they're different, it's cool! At quick glance, they looked the same! Disregard this comment! :)

PR #134, opened 2017-12-07
VAM-130: ORWU VERSRV

creindle:
[X] Has correct files
[X] Passes comparison tests


emulation/models/orwu-versrv.js
```diff
@@ -0,0 +1,46 @@
...
+const cprsMetaService = require('../../services/cprsMetaService');
+const _ = require('lodash');
+
+const RPC_EMULATOR_MODEL = [
+    {
+        id: 'ORWU_VERSRV',
+        name: 'ORWU VERSRV',
+        function: 'VERSRV^ORWU',
+        domain: 'CPRS',
+        description: 'return server version of option name. This is specifically used by CPRS GUI to determine the'
```
mfuroyama: Let's get rid of the string concatenation here for the string constant, and move the string onto one line. Actually, I'm surprised the ESLinter didn't catch this, since it's part of the AirBnB JS linting spec:

https://github.com/airbnb/javascript#strings--line-length

services/cprsMetaService/lookupServerVersion.js
```diff
@@ -0,0 +1,35 @@
...
+function lookupServerVersion(optionName, context, callback) {
+    const upperCaseOptionName = optionName.toUpperCase();
+    this.find('Option', {
+        query: { label: upperCaseOptionName, localStationNumber: context.stationId },
+        projection: serviceUtils.createProjection(['menuText']),
+    }).then((res) => {
+        if (!res || _.isEmpty(res)) return callback(null, '0.0.0.0');
+        const menuText = _.get(res[0], 'menuText', '');
+        let version = menuText.split('version')[1] || '';
+        if (version) version = version.trim();
```
mfuroyama: I think you probably don't need the check for `version` here, since the way you have the assignment setup in the previous line, `version` will *always* be a string, right?

creindle: Do you mean Line 25? Yes, version will ALWAYS be a string.

PR #135, opened 2017-12-08
VAM-54: ORWDBA1 BASTATUS

creindle:
[X] Passes comparison tests
[X] Created necessary files


emulation/models/tests/orwdba-1-bastatus-spec.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
+
+const rpcEmulatorModel = require('../orwdba-1-bastatus').rpcEmulatorModel;
+const unitTestUtils = require('./unitTestUtils');
+
+describe('ORWDBA 1 BASTATUS asynchronous specs', () => {
```
mfuroyama: Ahh, I know this was probably auto-generated, but can we modify the RPC name in the `describe` description text to match the emulated RPC name? (ORWDBA1 BASTATUS)

services/cprsMetaService/isBillingAwarenessEnabled.js
```diff
@@ -0,0 +1,31 @@
...
+ * @return {Boolean}   isBaEnabled  OR BILLING AWARENESS STATUS value field true or false
+ *
+ * @memberof CprsMetaService
+ * @instance
+ */
+function isBillingAwarenessEnabled(context, callback) {
+    this.find('Install', {
+        query: { label: 'PX CLINICAL INDICATOR DATA CAPTURE 1.0', localStationNumber: context.stationId },
+    }).then((res) => {
+        if (!res || _.isEmpty(res)) return callback(null, false);
```
mfuroyama: I think this code doesn't behave like you think it should behave...

In your code here, if `res` is say, empty, you call `return callback(...)`. It looks like you're expecting to be returned out of the entire function. **But** in reality, what happens next is the Promise chain will wrap the return value (undefined) in a promise and pass it to the next `then` handler, which will call the `callback` function again. In that case `callback` is called twice, which is a **not** something we want to do.

Test it out by:
1) adding a sample `console.log()` statement in the second `then` handler
2) call `isBillingAwarenessEnabled` with a junk station ID.

You'll see that your test `console.log()` statement is **always** called, success or failure.

Let me know if you you have any questions about this, or if this is confusing...

creindle: Yeah this was definitely a little bit confusing! I had to ask @kekuso about it! But I'll go through and retest the code out like how you say to do!

creindle: So, should I should simply have a `return false` instead?

PR #137, opened 2017-12-08
Vam 122 orwu clinloc

kekuso:
- create orwu-clinloc model, service, and tests
- move ORIMO ISCLOC, ORWDRA32 LOCTYPE, ORWPCE1 NONCOUNT services to locationMetaService


dataAccess/mongoDB/schemas/WardLocation.json
```diff
@@ -0,0 +1,207 @@
...
+                    "type": "STRING"
+                }
+            }
+        ],
+        "glHeader": {
+            "type": "STRING"
+        },
+        "nserv": {
+            "type": "STRING",
+            "enum": []
```
mfuroyama: Hmm, for some reason, the conversion script missed the `enum` fields here: Can we add the following (per the Fileman schema):
```
MED
SUR
ICU
PSY
```

Thanks!

services/locationMetaService/lookupClinicList.js
```diff
@@ -0,0 +1,141 @@
...
+ * @param  {Object}   context  Session context information associated with the service call.
+ * @param  {Function} callback Function called on process completion, with standard (err, result) signature.
+ * @return {Array} Array of Location Objects with label and id properties.
+ *
+ * @memberof LocationMetaService
+ * @instance
+ */
+function lookupClinicList(searchInput, direction, context, callback) {
+    if (direction.toString() !== '-1' && direction.toString() !== '1') {
+        return callback(null, new Error('second parameter must be 1 or -1'));
```
sidjtd: Hey, in your model file, in invokeRPC, I noticed you did a really cool typecheck where you make anything that isn't -1 default to 1, whereas the original it only did that with an empty string. ('' is converted to 1).

Since you did that, only 1 and -1 should be passed into the service function... right? Isn't this not "Trusting the contract of the Layer Below"?  If I'm not seeing the whole picture, LMK!!!

kekuso: 👍

PR #138, opened 2017-12-09
Vam 131 lookup new location

sidjtd:
Here is the PR. Did not expect those last hour turn of events....


services/locationMetaService/lookupNewLocationList.js
```diff
@@ -0,0 +1,124 @@
...
+];
+
+const WARD_PROJECTION = [
+    'id',
+    'outofserviceDate',
+];
+
+const today = moment().format('YYYY-MM-DD');
+
+function sortList(searchTerm, dir, locList) {
```
mfuroyama: This is okay for now, but going forward, we'll see if we can somehow get sorting done through the Data Access / MongoDB built-in mechanisms

services/locationMetaService/lookupNewLocationList.js
```diff
@@ -0,0 +1,124 @@
...
+ * @return {Array} Array of Location Objects with label and id properties.
+ *
+ * @memberof LocationMetaService
+ * @instance
+ */
+function lookupNewLocationList(searchInput, direction, context, callback) {
+    const service = this;
+    let hospitalList = [];
+
+    service.find('HospitalLocation', { query: { localStationNumber: 6100 }, projection: serviceUtils.createProjection(HOSPITAL_PROJECTION) })
```
mfuroyama: Hmm...let's avoid hard-coding the station identifiers here. The information should be in the `context` object

PR #141, opened 2017-12-11
VAM-308: Enhance Model Validity Checker

creindle:
[X] Ensure that the `complexity` field string only contains one of the acceptable, enumerated values (LOW, ~MODERATE~ MEDIUM (Per @mfuroyama ), HIGH)
[X] Ensure that other enumerated field strings (e.g. `rpcType`, `returnType`, `inputParameters.parameterType`) only contains one of the acceptable listed values.
[X] Ensure that there are no stray, misnamed fields. The RPC objects should only have fields as specified in the RPC Model Object Fields spec (RPC models should only have fields listed for All and VICS systems)


emulation/models/tests/models-spec.js
```diff
@@ -47,10 +47,17 @@ describe('rpcEmulatorModel', () => {
...
         let objId;
 
-        VISTA_STACK_KEYS.forEach((key) => {
-            it(`${objInRPC.id} required field "${key}" should be present`, () => {
-                expect(_.includes(objectKeys, key)).toBe(true);
-            });
+        it(`${objInRPC.id}'s keys should match model`, () => {
+            const differenceInKeys = _.xor(VISTA_STACK_KEYS, objectKeys);
+            if (differenceInKeys.length === 1) {
+                expect(_.includes(differenceInKeys, 'validateParams') || _.includes(differenceInKeys, 'formatResult')).toEqual(true);
```
creindle: Can you let me know if there's a better way to check for the optional 'validateParams' and 'formatResult'?

mfuroyama: Ahh, okay! You might want to consider having *two* sets of object keys: a `REQUIRED_RPC_MODEL_KEYS` list and an `OPTIONAL_RPC_MODEL_KEYS` list, then check in two stages:  First, check to see that all the required fields are present:
```Javascript
const differenceInKeys = _.xor(REQUIRED_RPC_MODEL_KEYS, objectKeys);
```

Then, compare the `differenceInKeys` to the optional list, and generate errors accordingly. Keep in mind that the checker should ensure that the RPC Models:
1) **Shall** contain **all** the required fields
2) **May** contain one or more of the optional fields
3) **Shall NOT** contain any fields that aren't in the list

creindle: Thank you!

PR #143, opened 2017-12-11
VAM-299: List Exam Type lacks a "status" filter

creindle:
[X] Added inactiveFlag filter


services/pceMetaService/lookupExamType.js
```diff
@@ -15,15 +15,14 @@ const _ = require('lodash');
 function lookupExamType(session, callback) {
     /* Get all active Exam IDs and labels */
     this.find('Exam', {
-        query: { localStationNumber: { $eq: session.stationId } },
+        query: { inactiveFlag: { $exists: false }, localStationNumber: { $eq: session.stationId } },
```
mfuroyama: Actually, consider reworking the query object to remove the `$eq`:
```Javascript
query: { inactiveFlag: { $exists: false }, localStationNumber: session.stationId }
```

creindle: :O :O Yes! I apologize, I should've noticed this!

services/pceMetaService/lookupExamType.js
```diff
@@ -15,15 +15,14 @@ const _ = require('lodash');
...
         projection: serviceUtils.createProjection(['label', 'id']),
     })
         .then((examArr) => {
-
-            let elementLabel = element => element.label;
-            let element = el => ({ id: serviceUtils.getIdFromUrn(el.id), label: el.label }); 
-            const res = _(examArr).sortBy(elementLabel).map(element).value();           
+            const elementLabel = element => element.label;
+            const element = el => ({ id: serviceUtils.getIdFromUrn(el.id), label: el.label });
+            const res = _(examArr).sortBy(elementLabel).map(element).value();
```
mfuroyama: Two things here:
1) Nifty little lodash trick: If you're sorting an array of objects by a field, you can specify that field name as the sort iteratee. So, for example, the above becomes:
```Javascript
const res = _(examArr).sortBy('label').map(element).value();
```
You don't even need the anonymous `elementLabel` function

2) Hmm...I'm noticing here that the Service function is converting the URN IDs to IENs in the results. Could I ask you to do a solid and (a) remove the `map` conversion here and (b) modify the `formatResult` function of the calling RPC to do the conversion? Thanks!! 🙏

creindle:
Wait a second, I'm fixing the merge

services/pceMetaService/listExamTypes.js
```diff
@@ -15,17 +15,18 @@ const _ = require('lodash');
...
+        query: { inactiveFlag: { $exists: false }, localStationNumber: session.stationId },
+        // Parsing the query to only retrieve items from two keys
         projection: serviceUtils.createProjection(['label', 'id']),
     })
         .then((examArr) => {
-            // Sort the list alphabetically by label
-            const elementLabel = element => element.label;
-            const element = el => ({ id: serviceUtils.getIdFromUrn(el.id), label: el.label });
-            const res = _(examArr).sortBy(elementLabel).map(element).value();
+            const element = el => ({ id: el.id, label: el.label });
```
mfuroyama: This is fine as-is, but here's another cool lodash trick:

The `element` function above is the same as `_.pick(el, 'id', 'label');`

`_.pick` is one of my favorite lodash functions: https://lodash.com/docs/4.17.4#pick

PR #145, opened 2017-12-11
VAM-307: Remove service, stationNo lookup in askPCE

creindle:
[X] Added changes based on additional comments


mfuroyama:
After looking at this code, I realize there are several other things that we should probably fix that weren't in the original issue. I'll include the additional issues in the Jira issue

PR #146, opened 2017-12-11
Vam 297 change pce service lookups to list

kekuso:
change service names from "lookup____" to "list_____" when a list of data is returned


services/pceMetaService/listHealthFactorTypes.js
```diff
@@ -7,35 +7,32 @@ const _ = require('lodash');
...
-                    categoryId: serviceUtils.getIdFromUrn(_.get(el, 'category.id', '')),
-                });
+            const elementFactor = element => element.factor;
+            const element = el => ({
+                id: serviceUtils.getIdFromUrn(el.id),
+                factor: el.factor,
+                entryType: el.entryType[0],
+                categoryId: serviceUtils.getIdFromUrn(_.get(el, 'category.id', '')),
+            });
             const res = _(heathFactorArr).sortBy(elementFactor).map(element).value();
```
mfuroyama: Here's a neat little lodash trick: If you're using `sortBy` to an array of objects, and you're sorting solely on the contents of a particular field in the objects, you can just pass the name of that field in as the "sort" function argument:
```Javascript
const res = _(heathFactorArr).sortBy('factor').map(element).value();
```

kekuso: cool!

PR #152, opened 2017-12-12
Vam 298 include category boolean for list health factors

kekuso:
- pass Boolean param 'includeCategories' to listHealthFactorTypes service
- fix test specs


emulation/models/orwpce-get-health-factors-ty.js
```diff
@@ -26,11 +26,12 @@ const RPC_EMULATOR_MODEL = [
             example: '1',
         }],
         invokeRPC(rpcArgs, session, callback) {
-            pceMetaService.listHealthFactorTypes(rpcArgs, session, callback);
+            const includeCategories = rpcArgs ? !!rpcArgs : false;
```
mfuroyama: I think the ternary check here might be redundant, consider refactoring that out

PR #160, opened 2017-12-13
Misc fixes

sidjtd:
lookupAllergySiteParam already existed, seems getAllergySiteParam just wasn't deleted.

Shrunk the vertical height of some of the code where possible but no real change to how the code works......


mfuroyama:
Thanks for doing this, @sidjtd!

PR #162, opened 2017-12-19
VAM-306: Added check for incorrect queries in entArr

creindle:
Please let me know if there's a better way to perform this check!


services/parameterService.js
```diff
@@ -189,6 +189,14 @@ class ParameterService extends BaseService {
...
         const callback = _.isFunction(cb) ? cb : _.noop;
         const service = this;
+        const entityTypeMapKeys = _.keys(ENTITY_TYPE_MAP);
+        const differenceOfEntities = _.xor(opts.entity, entityTypeMapKeys);
+        _.each(differenceOfEntities, (queryInEntArr) => {
+            const isAKey = _.includes(entityTypeMapKeys, queryInEntArr);
+            if (!isAKey && !_.includes(queryInEntArr, `${context.stationId}:${ENTITY_MAP.LOC.fileType}`)) {
+                console.log(`Incorrect query ${queryInEntArr} in entity arr ${name}. Parameter Service should ONLY take a concrete id for LOCATION (44)`);
+            }
+        });
```
mfuroyama: Several things:
1. First of all, remember the intent of the issue is to prevent invalid entities from being passed into the `get` call by raising/returning an error. This code block doesn't do this, it merely pushes an error message out to standard out if an error case occurs.
2. Let's try not to leave `console.log` calls in our code base
3. It's probably better here to do the check against `options.entity` (as opposed to the `opts.entity`) since code using `options` object created within the function will be better guarded against edge cases.
4. Remember that the `entity` field can either be an array **or** a string

There are definitely better ways to do the checks, but let's address these hurdles first

creindle: Ok! I apologize for leaving the `console.log` in the code base, I wasn't too sure what to do at the point an error is found and wanted to discuss that with you. I know we're not able to return `fail()` in the service functions.

services/parameterService.js
```diff
@@ -189,7 +189,25 @@ class ParameterService extends BaseService {
         const options = _.defaults({}, opts, { entity: 'ALL' });
         const callback = _.isFunction(cb) ? cb : _.noop;
         const service = this;
+        const entityTypeMapKeys = _.keys(ENTITY_TYPE_MAP);
+        let error;
+        if (_.isArray(options.entity)) {
+            const differenceOfEntities = _.xor(options.entity, entityTypeMapKeys);
```
kekuso: I think you can reduce this block to a single check :)
`if(_.includes(entityTypeMapKeys, options.entity)) error = ...`

creindle: @kekuso Why isn't it posssssibleeeee!???

kekuso: Hmm maybe not a single check...forgot the values could be random URN's

emulation/models/orwpce-askpce.js
```diff
@@ -36,7 +36,7 @@ const RPC_EMULATOR_MODEL = [
 
         invokeRPC: (rpcArgs, session, callback) => {
             const locationURN = rpcArgs[1] ? modelUtils.createURNFromParams('44', rpcArgs[1], session.stationId) : '';
-            const entArr = ['USR', locationURN, 'SRV', 'DIV', 'SYS', 'PKG'];
+            const entArr = ['USR', locationURN, 'haha', 'DIV', 'SYS', 'PKG'];
```
mfuroyama: 😒  Umm...very funny I suppose, but whut? Why even change this?

emulation/models/tests/orwpce-askpce-spec.js
```diff
@@ -16,7 +16,7 @@ describe('ORWPCE ASKPCE asynchronous specs', () => {
     });
 
     it('invokeRPC: Integer arg returns integer', (done) => {
-        rpc.run(52, context, (err, res) => {
+        rpc.run([62, 4], context, (err, res) => {
```
mfuroyama: Same here? Is there a reason this was changed out-of-context?

services/parameterService.js
```diff
@@ -189,7 +189,26 @@ class ParameterService extends BaseService {
         const options = _.defaults({}, opts, { entity: 'ALL' });
         const callback = _.isFunction(cb) ? cb : _.noop;
         const service = this;
+        const entityTypeMapKeys = _.keys(ENTITY_TYPE_MAP);
```
mfuroyama: This array of keys is pretty constant, and doing this in the `get` function means that `_.keys` will be called and this array will be built **every time** the function gets called.

As a mini-optimization, let's move the generation of the key array outside the scope of this function and move it near to where `ENTITY_TYPE_MAP` is generated.

emulation/models/orwpce-askpce.js
creindle: I apologize! I was testing the new features I was adding to the parameterService :O

services/parameterService.js
```diff
@@ -189,7 +189,26 @@ class ParameterService extends BaseService {
...
+            });
+        }
+        if (_.isString(options.entity) && !_.includes(entityTypeMapKeys, options.entity)) {
+            error = `'${options.entity}' is an incorrect query as an entity`;
+        }
         return new Promise((resolve, reject) => {
+            if (error) {
+                callback(error);
+                reject(error);
+            }
```
mfuroyama: This is okay, but here's an alternate implementation to consider:
1. Consider defining a function, say, `isValidEntity` outside the scope of `get` that checks if `entities` passed in is valid:
   * If `entities` is an array, use `_.every` with recursion back into `isValidEntity`
   * Otherwise, just run the check (`isString && (ENTITY_MAP[entity] || isLocationURN));

2. Call that function within the context of the Promise function

This might not make sense...let's discuss

services/parameterService.js
```diff
@@ -41,6 +41,16 @@ const ENTITY_FILETYPE_KEY_MAP = _.transform(ENTITY_MAP, (map, value, key) => {
 const PARAMETER_FIELDS = ['id', 'instance', 'value', 'entity', 'wordProcessingText', 'parameter'];
 const PARAM_DEF_FIELDS = ['id', 'label', 'allowableEntities', 'localStationNumber'];
 
+const isValidEntity = (entities) => {
+    if (_.isArray(entities)) {
+        return _.every(entities, isValidEntity);
+    }
+    if (_.isString(entities) && (ENTITY_MAP[entities] || (serviceUtils.getFileTypeFromId(entities) === '44') || entities === 'ALL')) {
```
mfuroyama: Consider removing the if/else and just returning the calculated value above

creindle: That's cool it works simply like that!

PR #165, opened 2017-12-20
Vam 331 emulation must check rpc arguments consistently

kekuso:
- update gitignore file
- add more type validation to pureEmulation validation checks
- option for default validation checks to be made more strict by validateParams() in models
- make parameterTypes uniform (use 'NUMBER' instead of 'NUMERIC')


emulation/dispatchers/pureEmulation.js
```diff
@@ -111,11 +116,23 @@ class PureEmulationDispatcher {
...
-
             if (rpcModel.inputParameters.length >= rpcArgs.length) {
                 const type = rpcModel.inputParameters[index].parameterType;
-                if (((type === 'NUMBER') || (type === 'NUMERIC')) && !_.isNumber(+arg)) {
-                    error = `RPC call ${rpcModel.id}: Argument ${index} should be numeric [${rpcArgs}]`;
+                if (type === 'NUMBER' && !_.isNumber(+arg)) {
+                    error = `RPC call ${rpcModel.id}: Argument ${index} should be NUMBER [${rpcArgs}]`;
+                    return true;
+                }
+                if (type === 'LITERAL' && !_.isNumber(arg) && typeof arg !== 'string') {
```
mfuroyama: Ahh...I hate to ask this, but could we change `LITERAL` to `STRING` to match the [RPC Model Object Field spec](https://github.com/vistadataproject/VICSServer/wiki/RPC-Model-Object-Fields)?

The reason I ask is because this means we'll have to change the rest of the RPC models... :/ Let's discuss if you like...

emulation/dispatchers/pureEmulation.js
```diff
@@ -29,12 +29,17 @@ class PureEmulationDispatcher {
...
+        const parameterValidator = PureEmulationDispatcher.validateModelParams;
+        const customParamValidator = model.validateParams || '';
         const validationResult = parameterValidator(model, args);
+        const customValidationResult = customParamValidator ? customParamValidator(model, args) : '';
         if (validationResult) {
             return cb(new Error(validationResult));
         }
+        if (customValidationResult) {
+            return cb(new Error(customValidationResult));
+        }
```
mfuroyama: Also, since we're changing the validation behavior described in the [RPC Model object spec](https://github.com/vistadataproject/VICSServer/wiki/RPC-Model-Object-Fields), let's change the description for the `validateParams` field. Thanks!

emulation/dispatchers/pureEmulation.js
kekuso: Hmm so should I change all the RPC models' inputParameter.parameterType values from 'LITERAL' to 'STRING'?

emulation/dispatchers/pureEmulation.js
kekuso: 🆗

PR #161, opened 2017-12-18
VAM-334: Rename Service.autoVisitType to "getAutoVisitType"

creindle:

PR #167, opened 2017-12-21
VAM-355: Add service function names to "services" field arrays in RPC Emulation Models

creindle:


emulation/models/orevntx-1-dlgien.js
```diff
@@ -18,7 +18,7 @@ const RPC_EMULATOR_MODEL = [
         files: ['101.41'],
         parameters: [],
         labels: ['META'],
-        services: ['utilityMetaService'],
+        services: ['orderMetaService.lookupOrderDialogId'],
```
creindle: I'm curious about adding this because in the `invokeRPC` method, it calls on the `orderMetaService`, rather than `utilityMetaService`. So should I leave both?

emulation/models/orwdx-dgnm.js
```diff
@@ -18,7 +18,7 @@ const RPC_EMULATOR_MODEL = [
         files: ['100.98'],
         parameters: [],
         labels: ['META'],
-        services: ['utilityMetaService'],
+        services: ['utilityMetaService', 'orderMetaService.lookupDisplayGroupId'],
```
creindle: This is in relation to the comment above, should I have both services in this field?

emulation/models/orevntx-1-dlgien.js
mfuroyama: No, just leave the `orderMetaService` listing, since the whole purpose of this exercise is to ensure that we're **correctly** documenting the service function use in the RPCs. 

This is good; you killed two birds with one stone here: You added the service function **and** you fixed an incorrect `services` listing! 👍

emulation/models/orwch-loadall.js
```diff
@@ -17,7 +17,7 @@ const RPC_EMULATOR_MODEL = [
         files: [],
         domain: 'ORDER',
         labels: ['META'],
-        services: ['orderMeta'],
+        services: ['parameterService'],
```
mfuroyama: 👍  for the fix. For these `parameterService` listings, let's add the `get` function to the string

emulation/models/orwdx-dgnm.js
mfuroyama: No, as mentioned in a comment above, the point of this is to **accurately** document the service function usage. Since we don't actually use the `utilityMetaService` (actually, it doesn't exist anymore, anyway) we can remove it here.

emulation/models/orwpce-auto-visit-type-select.js
```diff
@@ -14,7 +14,7 @@ const RPC_EMULATOR_MODEL = [
         files: [],
         domain: 'PCE',
         labels: ['META', 'VISIT'],
-        services: ['pceMetaService'],
+        services: ['pceMetaService.autoVisitType'],
```
mfuroyama: You **might** want to change this to match the modifications made in pull request #161:

`pceMetaService.autoVisitType => pceMetaService.getAutoVisitType`

emulation/models/orevntx-1-dlgien.js
creindle: You're welcome!

PR #168, opened 2017-12-22
VAM-464: Fix Implementation to prevent cross-user data access

creindle:


emulation/models/tiu-template-access-level.js
```diff
@@ -50,8 +50,9 @@ const RPC_EMULATOR_MODEL = [
         invokeRPC: (rpcArgs, session, callback) => {
             const template = rpcArgs[0] ? modelUtils.createURNFromParams(8927, rpcArgs[0], session.stationId) : '';
             const user = rpcArgs[1] ? modelUtils.createURNFromParams(200, rpcArgs[1], session.stationId) : '';
-            const location = rpcArgs[2] ? modelUtils.createURNFromParams(49, rpcArgs[2], session.stationId) : '';
-            documentMetaService.lookupTiuTemplateAccessLevel(template, user, location, session, callback);
+            if (user !== session.userId) return callback(new Error('Security Risk: User IEN from Client does not match with session user IEN'));
```
mfuroyama: Ahh, we can leave out the 'Security Risk' here... :)

services/documentMetaService/lookupTiuTemplateAccessLevel.js
```diff
@@ -13,19 +13,18 @@ const TEMPLATE_ACCESS_LEVEL = [
 
 /**
  * @summary Returns Template Access level of User (0 = FULL ACCESS, 1 = READ ONLY, 2 = NO ACCESS,3 = SHARED TEMPLATES EDITOR - ACCESS PARAMETERS DO NOT APPLY
- * @param  {Number}   template The template ID to check for user's access level.
- * @param {Number} user A specified user ID.
- * @param {Number} location A specified location ID (optional).
+ * @param  {Number}   template  The template ID to check for user's access level.
+ * @param  {Number}   location  A specified location ID (optional).
```
mfuroyama: Minor change, but let's update the JSDoc parameter types for `template` and `location` to `{String}`. These should be URNs and we want our API docs to represent correctly. We might also want to add something about how these values are now URNs. 👍

services/documentMetaService/lookupTiuTemplateAccessLevel.js
```diff
@@ -35,15 +34,15 @@ function lookupTiuTemplateAccessLevel(template, user, location, context, callbac
         .then((tiuObj) => {
             const editorClass = _.get(tiuObj[0], 'editorClass.id', '');
             editorClassId = editorClass ? serviceUtils.getIdFromUrn(editorClass) : '';
-            return service.find('UsrClass', { query: { label: 'USER' }, projection: serviceUtils.createProjection(['id']) });
+            return service.find('UsrClass', { query: { label: 'USR' }, projection: serviceUtils.createProjection(['id']) });
```
mfuroyama: Ooh, actually, okay, this one should be changed back to `'USER'`, since this isn't a Parameter Service query

creindle: Oooooooooh! That was my mistake I apologize!

creindle:
Also, I added to my comment to #10 of the VICS Coding Guidelines in regards to fixing this issue.

PR #169, opened 2017-12-22
Adjusted entity inputs to correct entities

creindle:
Due to the changes in Parameter Service that only allows for either service keywords or a concrete location URN, adjustments had to be made for RPC models that were passing in incorrect entities. These errors were found when running the model tests.


emulation/models/orwpce-askpce.js
```diff
@@ -35,7 +35,7 @@ const RPC_EMULATOR_MODEL = [
         }],
 
         invokeRPC: (rpcArgs, session, callback) => {
-            const locationURN = rpcArgs[1] ? modelUtils.createURNFromParams('44', rpcArgs[1], session.stationId) : '';
+            const locationURN = rpcArgs[1] ? modelUtils.createURNFromParams('44', rpcArgs[1], session.stationId) : 'LOC';
```
creindle: Changes to the Parameter Service do not allow for empty strings to be passed in. Therefore this was changed to an relevant, accepted service keyword.

emulation/models/orwpce-always-checkout.js
```diff
@@ -30,9 +30,8 @@ const RPC_EMULATOR_MODEL = [
 
         invokeRPC: (rpcArgs, session, callback) => {
             let loc = rpcArgs[0] || 'LOC';
-            loc = (!_.isNaN(parseFloat(loc))) ? modelUtils.createURNFromParams(44, loc, session.stationId) : loc;
+            loc = (_.isNumber(loc)) ? modelUtils.createURNFromParams(44, loc, session.stationId) : 'LOC';
```
creindle: The original check did not catch letter inputs, which gave an error from the Parameter Service due to it only accepting service keywords or location URNs.

mfuroyama: Wait, what? :)

Anyway, the first argument might be a number (e.g. 1) or a string containing a number (e.g. '1') so we'll need to account for those two cases.

One way to do this would be to use the numeric casting operator (`+`) to convert the input argument, then check for `!_.isNaN`:

```Javascript
const loc = !_.isNaN(+rpcArgs[0]) ? modelUtils.createURNFromParams(44, rpcArgs[0], session.stationId) : 'LOC';
```

emulation/models/orwpce-askpce.js
mfuroyama: LOL! The change above should mirror this one! Best to do the numeric casting, just in case:

```Javascript
const locationURN = +rpcArgs[1] ? modelUtils.createURNFromParams('44', rpcArgs[1], session.stationId) : 'LOC';
```

creindle: Perfect thank you!

emulation/models/orwpce-always-checkout.js
creindle: Perfecccccttt!!! Thank you!

creindle: And! To explain what happened, the check before would allow inputs such as "b" to be passed in as `loc` and a URN would be created from this input! So what you wrote is better!!

PR #173, opened 2017-12-26
Created Location Entity Utility

creindle:
[X] Services with new utility passes comparison tests
[X] Models with new utility passes model tests


emulation/models/orimo-iscloc.js
```diff
@@ -3,7 +3,7 @@
 'use strict';
 
 const locationMetaService = require('../../services/locationMetaService');
-const modelUtils = require('./modelUtils');
+const serviceUtils = require('../../services/serviceUtils');
```
kekuso: Consider importing modelUtils instead of serviceUtils. I think it would be good to create the same function (createLocationEntity) in modelUtils to promote segregation of services/models (yay segregation!)

creindle: Ooooh! I understand what you mean! I'll definitely do that! 
Thank you Adam!

creindle: Also, in another branch, I'll correct the files to ensure the segregation is strict.

emulation/models/orwpce-get-excluded.js
```diff
@@ -41,7 +41,7 @@ const RPC_EMULATOR_MODEL = [
         }],
 
         invokeRPC: (rpcArgs, session, callback) => {
-            const locationURN = parseInt(rpcArgs[0], 10) ? modelUtils.createURNFromParams('44', rpcArgs[0], session.stationId) : session.locationId;
+            const locationURN = modelUtils.createLocationEntity(rpcArgs[0], session);
```
creindle: I realized that if `rpcArgs[0]` is an incorrect input, this would return "LOC", which I believe is not a wanted output.

services/serviceUtils.js
```diff
@@ -109,6 +109,14 @@ const createProjection = (projections) => {
 
 const createSimpleQuery = query => ({ query });
 
+/**
+ * @description Checks if input is correct format to create location URN. If not, returns acceptable entity 'LOC'
+ * @param  {Number|String} locationId - location ID used to construct location URN
+ * @return {Object} session - User context containing station ID used to construct location URN
```
mfuroyama: Same here: let's update the description of the `return` value here in the JSDoc to more accurately describe what's actually being returned by the function?

emulation/models/orwpce-get-excluded.js
mfuroyama: Yeah, you're right. In this case, consider replacing the `createURNFromParams` with your utility function. The line might end up looking like this:
```Javascript
const locationURN = parseInt(rpcArgs[0], 10) ? modelUtils.createLocationEntity(rpcArgs[0], session) : session.locationId;
```

emulation/models/orwdx-wrlst.js
```diff
@@ -48,7 +48,7 @@ const RPC_EMULATOR_MODEL = [
         }],
 
         invokeRPC: (rpcArgs, session, callback) => {
-            const location = rpcArgs[0] ? modelUtils.createURNFromParams('44', rpcArgs[0], session.stationId) : session.locationId;
+            const location = modelUtils.createLocationEntity(rpcArgs[0], session);
```
creindle: This needed to be fixed based on the code suggestion in `orwpce-get-excluded`

services/serviceUtils.js
```diff
@@ -109,6 +109,14 @@ const createProjection = (projections) => {
 
 const createSimpleQuery = query => ({ query });
 
+/**
+ * @description Checks if input is correct format to create location URN. If not, returns acceptable entity 'LOC'
+ * @param  {Number|String} locationId - location ID used to construct location URN
+ * @return {String} session - Location URN or acceptable entity 'LOC'
```
mfuroyama: Let's add the `session` as a `@param` with `{Object}` type here in the JSDoc comment

emulation/models/modelUtils.js
```diff
@@ -120,6 +120,14 @@ const isEmptyString = str => (_.isString(str) && _.isEmpty(str));
  */
 const isNilOrFalse = val => (_.isNil(val) || (_.isBoolean(val) && !val));
 
+/**
+ * @description Checks if input is correct format to create location URN. If not, returns acceptable entity 'LOC'
+ * @param  {Number|String} locationId - location ID used to construct location URN
+ * @return {Object} session - User context containing station ID used to construct location URN
```
mfuroyama: Two things:
1. Let's add the `session` as a `@param` with `{Object}` type here in the JSDoc comment
2. Let's also fix the `@return` comment here to match what's actually being returned.

services/serviceUtils.js
creindle: Ughhhhh I need to be more careful T.T

PR #174, opened 2017-12-26
Vam 124 orwu haskey pure js

kekuso:
- did it forealz this time
- moved hasSecurityKey to userMetaService
- rolled orwu nphaskey into this PR, since it utilizes the hasSecurityKey service


services/userMetaService/hasSecurityKey.js
```diff
@@ -0,0 +1,37 @@
...
+ *                             If a person URN is not provided, the session.userId URN will be used by default.
+ * @param  {String}   key      The name of the security key (i.e. 'XUPROGMODE')
+ * @param  {Object}   context  Session context information associated with the service call.
+ * @param  {Function} callback Function called on process completion, with standard (err, result) signature.
+ * @return {Boolean}
+ *
+ * @memberof UserMetaService
+ * @instance
+ */
+function hasSecurityKey(person, key, context, callback) {
```
mfuroyama: Let's remove the `person` parameter here and not even allow users of this function to even try passing this in. The comparison check against the person value passed in should happen in the RPC model, not the service.

services/userMetaService/hasSecurityKey.js
```diff
@@ -0,0 +1,37 @@
+'use strict';
+
+const serviceUtils = require('../serviceUtils');
+const _ = require('lodash');
+
+/**
+ * @summary Returns true if a user holds a security key, otherwise false.
+ * @param  {String}   person   The urn of the person to check for the security key (optional).
```
mfuroyama: Let's remove the `person` parameter here...

emulation/models/orwu-nphaskey.js
```diff
@@ -0,0 +1,45 @@
...
+        },
+        {
+            id: 'KEY',
+            parameterType: 'LITERAL',
+            required: true,
+            description: 'Key label',
+            example: 'PROVIDER',
+        }],
+        invokeRPC: (rpcArgs, session, callback) => {
+            const person = rpcArgs[0] ? modelUtils.createURNFromParams('200', rpcArgs[0], session.stationId) : '';
```
mfuroyama: ...same here, let's do the comparison check between `rpcArgs[0]` (which is supposed to be a user IEN) and the `session.userId` here in the RPC model and not in the service.

PR #177, opened 2017-12-27
VAM-411: ORWRP GET DEFAULT PRINTER

creindle:
[X] Created JSON for 'Device' collection to be added to mongoDB
[X] Created model, service, and spec files without RPC Emulation Model Template Generator as the model file was not created in the nonClinicalRPC directory
[X] Changed captured userId data to match session user id (63)
[X] Ensured user is not allowed to chose a userId different from their session userId
[X] Passes comparison and model tests

Note: This is the first RPC to be translated completely from MUMPS code, so please be stringent when checking if I implemented this incorrectly!


emulation/models/orwrp-get-default-printer.js
```diff
@@ -0,0 +1,49 @@
...
+    {
+        id: 'LOCATION',
+        parameterType: 'NUMBER',
+        required: false,
+        description: 'the user\'s location',
+        example: '0',
+    }],
+    invokeRPC(rpcArgs, session, callback) {
+        const user = rpcArgs[0] ? modelUtils.createURNFromParams(200, rpcArgs[0], session.stationId) : '';
+        if (user !== session.userId) return callback(new Error('User IEN from Client does not match with session user IEN'));
```
mfuroyama: 👍

emulation/models/orwrp-get-default-printer.js
```diff
@@ -0,0 +1,49 @@
...
+        const location = rpcArgs[1] ? modelUtils.createURNFromParams(44, rpcArgs[1], session.stationId) : 'LOC'; // To be replaced with location utility
+        orderMetaService.lookupDefaultPrinter(location, session, callback);
+    },
+    formatResult(res) {
+        if (!res) return '';
+        if (res === 'WIN;Windows Printer') return res;
+        if (_.isArray(res)) return `${modelUtils.getIenFromId(res[0])};${res[1]}`;
```
mfuroyama: Consider the following changes:
1. Adding a check for the `length` of `res`. If I'm not mistaken, the service function *might* return an array of length 1, which will cause this function to return invalid results.
2. Moving the array guard checks to the top of the function. This kills to birds with one stone: better error handling *and* best-practices clean execution path. Normally, if you have a function that is returning a value, it's good form to make sure **all** execution paths return a concrete value. Not that this will happen in this case, but let's say for example that `res` was say, a number. Example at the execution flow and see how this function would fall through to the end, and this function returns an `undefined` value. Bottom line: it's usually always better to end a function that returns a value with a non-conditional return statement.

services/orderMetaService/lookupDefaultPrinter.js
```diff
@@ -0,0 +1,49 @@
+'use strict';
+
+const _ = require('lodash');
+const ParameterService = require('../parameterService');
+const serviceUtils = require('../serviceUtils');
+
+/**
+ * @summary Returns default printer for user
+ * @param  {String}   labelArg  String representing name of data in Device Collection
```
mfuroyama: I think you can remove this parameter from the JSDoc comment

services/orderMetaService/lookupDefaultPrinter.js
```diff
@@ -0,0 +1,49 @@
...
+ * @param  {Function}   callback  Function called on process completion, with standard (err, result) signature.
+ * @return {Array}   aggregate  Array containing value from ORWDP WINPRINT DEFAULT and label from query
+ *
+ * @memberof OrderMetaService
+ * @instance
+ */
+function lookupDefaultPrinter(location, context, callback) {
+    const service = this;
+    const aggregate = [];
+    ParameterService.withContext(context).get('ORWDP WINPRINT DEFAULT', { entity: [location, 'ALL'] })
```
mfuroyama: You might want to consider only adding the `location` parameter to the array if it's defined, since the comments say that the parameter is *optional* and might not be included.

emulation/models/orwrp-get-default-printer.js
```diff
@@ -0,0 +1,49 @@
...
+'use strict';
+
+const _ = require('lodash');
+const orderMetaService = require('../../services/orderMetaService');
+const modelUtils = require('./modelUtils');
+
+const RPC_EMULATOR_MODEL = [{
+    id: 'ORWRP_GET_DEFAULT_PRINTER',
+    name: 'ORWRP GET DEFAULT PRINTER',
+    function: 'GETDFPRT',
```
mfuroyama: I'm pretty sure this is `GETDFPRT^ORWRP`. Let's change this to match what's listed as the corresponding MUMPS function in the Jira issue

services/orderMetaService/lookupDefaultPrinter.js
creindle: Ooooh! Good catch! Thank you for understanding that!

emulation/models/orwrp-get-default-printer.js
creindle: Awesome, thank you for the clarification!

emulation/models/orwrp-get-default-printer.js
creindle: Ah! I understand. Thank you for the explanation!

services/orderMetaService/lookupDefaultPrinter.js
```diff
@@ -0,0 +1,49 @@
...
+    const service = this;
+    const aggregate = [];
+    const entityArr = location ? [location, 'ALL'] : ['ALL'];
+    ParameterService.withContext(context).get('ORWDP WINPRINT DEFAULT', { entity: entityArr })
+        .then((res) => {
+            if (!res || _.isEmpty(res)) return false;
+            const result = _.get(res[0], 'value', '');
+            if (+result) return 'WIN;Windows Printer';
+            return ParameterService.withContext(context).get('ORWDP DEFAULT PRINTER', { entity: [location, 'ALL'] });
+        }).then((res) => {
```
mfuroyama: Really, really minor but let's make the position of this `then` call consistent with the rest of the chain

creindle: Ah! I apologize, I should've caught that!

PR #178, opened 2017-12-28
VAM-413: ORWU OVERDL

creindle:


emulation/models/orwu-overdl.js
```diff
@@ -0,0 +1,32 @@
...
+const ParameterService = require('../../services/parameterService');
+
+const RPC_EMULATOR_MODEL = [{
+    id: 'ORWU_OVERDL',
+    name: 'ORWU OVERDL',
+    function: 'OVERDL',
+    domain: '',
+    description: 'Return parameter value of ORPARAM OVER DATELINE',
+    returnType: 'SINGLE VALUE',
+    rpcType: '',
```
sidjtd: Hmmm, I think this rpcType is READ, domain is.... CPRS? Anyone can confirm or deny?

creindle: Thank you! I'll wait for feedback from others too 😭 I wrote this from scratch so I had no idea!

emulation/models/orwu-overdl.js
```diff
@@ -0,0 +1,32 @@
...
+
+'use strict';
+
+const _ = require('lodash');
+const ParameterService = require('../../services/parameterService');
+
+const RPC_EMULATOR_MODEL = [{
+    id: 'ORWU_OVERDL',
+    name: 'ORWU OVERDL',
+    function: 'OVERDL',
```
mfuroyama: I'm pretty sure this is `OVERDL^ORWU`. Let's change this to match what's listed as the corresponding MUMPS function in the Jira issue

emulation/models/orwu-overdl.js
mfuroyama: Yep, @sidjtd is right: let's change this to `'READ'`

emulation/models/orwu-overdl.js
```diff
@@ -0,0 +1,32 @@
...
+    domain: '',
+    description: 'Return parameter value of ORPARAM OVER DATELINE',
+    returnType: 'SINGLE VALUE',
+    rpcType: '',
+    complexity: 'LOW',
+    files: [],
+    parameters: ['ORWU OVERDL'],
```
mfuroyama: This should be the name of the parameter(s) that we access through the parameter service (here: `ORPARAM OVER DATELINE`)

creindle: 😱 I need to check that the other models list the name of the parameter(s) accessed too!

emulation/models/orwu-overdl.js
```diff
@@ -0,0 +1,32 @@
...
+const ParameterService = require('../../services/parameterService');
+
+const RPC_EMULATOR_MODEL = [{
+    id: 'ORWU_OVERDL',
+    name: 'ORWU OVERDL',
+    function: 'OVERDL^ORWU',
+    domain: 'CPRS',
+    description: 'Return parameter value of ORPARAM OVER DATELINE',
+    returnType: 'READ',
+    rpcType: '',
```
mfuroyama: I think you might have changed the wrong value 😄 
`returnType` should be `SINGLE VALUE`, and `rpcType` should be `READ`

creindle: Noooooo... nooooooo....

PR #179, opened 2017-12-29
Vam 55 list treatment factor hints

sidjtd:
Got thrown off by the order of the data in the model/mumpsy results for a second, but a explicit sort request via ID fixed that up really quick!   Should we move that outside of the service function, though? Technically the sorting is not needed by the service Function... it's only done to match the \ rpcRunner results....


emulation/models/orwdba-3-hints.js
```diff
@@ -0,0 +1,40 @@
...
+        function: 'HINTS^ORWDBA3',
+        domain: 'ORDER',
+        description: 'Return HINTS for ORBA Treatment Factors - used by Delphi',
+        rpcType: 'READ',
+        complexity: 'MODERATE',
+        returnType: 'ARRAY',
+        files: ['9.2'],
+        parameters: [],
+        labels: ['META'],
+        services: ['orderMetaService'],
```
mfuroyama: Let's add the name of the service function called here:
```orderMetaService. listTreatmentFactorHints```

sidjtd: Got it!!!! (Happy 1st!!!!)

PR #180, opened 2017-12-29
VAM-123: ORWU EXTNAME & Renamed fileManService to VISTAMetaService

creindle:
[X] Completed RPC and passes both model and comparison tests
[X] Renamed directory `fileManService` and all instances to `VISTAMetaService`


emulation/models/orwu-extname.js
```diff
@@ -0,0 +1,52 @@
...
+            const ien = rpcArgs[0] || 0;
+            const fileNum = rpcArgs[1] || '';
+            const urn = modelUtils.createURNFromParams(fileNum, ien, session.stationId);
+            let fileName = fileUtils.lookupFileLabelFromId(fileNum);
+            fileName = _.replace(fileName, ' ', '');
+            if (ien === 0) return callback(null, fileName);
+            if (!fileName) return callback(new Error('Incorrect file name'));
+            cprsMetaService.lookupExternalFormOfPointer(urn, fileName, session, callback);
+        },
+        formatResult: res => res,
```
mfuroyama: This is not necessary, you can remove `formatResult` here if it doesn't do anything

emulation/models/orwu-extname.js
```diff
@@ -0,0 +1,52 @@
...
+            required: false,
+            description: 'File Number',
+            example: '1',
+        }],
+        invokeRPC: (rpcArgs, session, callback) => {
+            const ien = rpcArgs[0] || 0;
+            const fileNum = rpcArgs[1] || '';
+            const urn = modelUtils.createURNFromParams(fileNum, ien, session.stationId);
+            let fileName = fileUtils.lookupFileLabelFromId(fileNum);
+            fileName = _.replace(fileName, ' ', '');
```
mfuroyama: What's going on here?

emulation/models/orwu-extname.js
```diff
@@ -0,0 +1,52 @@
...
+            id: 'FN',
+            parameterType: 'LITERAL',
+            required: false,
+            description: 'File Number',
+            example: '1',
+        }],
+        invokeRPC: (rpcArgs, session, callback) => {
+            const ien = rpcArgs[0] || 0;
+            const fileNum = rpcArgs[1] || '';
+            const urn = modelUtils.createURNFromParams(fileNum, ien, session.stationId);
```
mfuroyama: Let's move this declaration to right before the `lookupExternalFormOfPointer` function call. It makes for slightly more readable code, plus, you don't need it unless you make the call so it might end up as extra processing.

emulation/models/orwu-extname.js
```diff
@@ -0,0 +1,52 @@
...
+            let fileName = fileUtils.lookupFileLabelFromId(fileNum);
+            fileName = _.replace(fileName, ' ', '');
+            if (ien === 0) return callback(null, fileName);
+            if (!fileName) return callback(new Error('Incorrect file name'));
```
mfuroyama: Probably makes more sense to move this line directly after the definition of `fileName`

Think about what this function is doing and consider re-organizing using the comments here

emulation/models/orwu-extname.js
creindle: I apologize for missing this!

emulation/models/orwu-extname.js
creindle: This is much more clean! Thank you.

emulation/models/orwu-extname.js
creindle: Ah! That's good to keep in mind! Thank you!

services/cprsMetaService/lookupExternalFormOfPointer.js
```diff
@@ -0,0 +1,29 @@
...
+ * @param  {String}   specificId  Id of document to search for.
+ * @param  {String}   fileName  File name used as collection name.
+ * @param  {Object}   context  Session context information associated with the service call.
+ * @param  {Function} callback  Function called on process completion, with standard (err, result) signature.
+ * @return {String}   response  Value of label from specific document
+ *
+ * @memberof CprsMetaService
+ * @instance
+ */
+function lookupExternalFormOfPointer(specificId, fileName, context, callback) {
```
mfuroyama: You know, I feel like this function (where you get a label from an ID URN via query) is generic enough where you could probably move it out of CPRS and into the Service Utilities or something? Let's discuss

emulation/models/orwu-extname.js
```diff
@@ -0,0 +1,50 @@
...
+        id: 'ORWU_EXTNAME',
+        name: 'ORWU EXTNAME',
+        function: 'EXTNAME^ORWU',
+        domain: 'CPRS',
+        description: 'Returns the external form of a pointer value given the IEN and filenumber.',
+        rpcType: 'UTILITY',
+        complexity: 'LOW',
+        returnType: 'SINGLE VALUE',
+        files: [],
+        parameters: ['cprsMetaService.lookupExternalFormOfPointer'],
```
mfuroyama: This should go into the `services` array and the `parameters` array should probably be empty here...

mfuroyama:
Okay, lets do deees:
1. Move the service function from CPRS to VISTA
2. Rename it to something more meaningful and appropriate (e.g. `getLabel`)
3. Change code that uses VISTAMetaService to do the following require:
`const VISTAService = require('../VISTAMetaService');`

emulation/models/orwu-extname.js
creindle: Yes, I think this may be the second time I made this mistake so I need to double check my work in other models...

emulation/models/orwu-extname.js
```diff
@@ -2,26 +2,23 @@
...
-const fileUtils = require('../../services/VISTAMetaService/fileUtils');
+const VISTAService = require('../../services/VISTAMetaService');
 
 const RPC_EMULATOR_MODEL = [
     {
         id: 'ORWU_EXTNAME',
         name: 'ORWU EXTNAME',
         function: 'EXTNAME^ORWU',
-        domain: 'CPRS',
+        domain: 'VISTA',
```
mfuroyama: Let's change this back to `CPRS` here. Even though we're using the VISTA meta service in the implementation, the RPC still belongs to the CPRS domain.

emulation/models/orwu-extname.js
```diff
@@ -2,26 +2,23 @@
...
         rpcType: 'UTILITY',
         complexity: 'LOW',
         returnType: 'SINGLE VALUE',
         files: [],
-        parameters: ['cprsMetaService.lookupExternalFormOfPointer'],
+        parameters: [],
         labels: ['META'],
-        services: [],
-        mvdm: {},
+        services: ['VISTAService.lookupFileLabelFromId', 'VISTAService.getLabel'],
```
mfuroyama: `VISTAService` => `VISTAMetaService` in the text here

services/VISTAMetaService/getLabel.js
```diff
@@ -3,17 +3,18 @@
...
  * @param  {Function} callback  Function called on process completion, with standard (err, result) signature.
  * @return {String}   response  Value of label from specific document
  *
- * @memberof CprsMetaService
+ * @memberof VISTAMetaService
  * @instance
  */
-function lookupExternalFormOfPointer(specificId, fileName, context, callback) {
+
+function getLabel(fileName, specificId, context, callback) {
```
mfuroyama: Really minor readability comment here: consider renaming the `specificId` parameter to something, um, more "specific". 😄  The JSDoc comments say that `specificId` is the "Id of document to search for", so a more focused variable name (like, say `documentId` for example) would help readers of this code not only understand the purpose of the parameter, but also better understand what the overall code is doing.

creindle: I apologize about that!

emulation/models/orwu-extname.js
creindle: Oh that's interesting!

PR #181, opened 2018-01-03
VAM-89: ORWPCE ACTIVE PROV

creindle:


emulation/models/orwpce-active-prov.js
```diff
@@ -0,0 +1,49 @@
...
+        }, {
+            id: 'ORWDT',
+            parameterType: 'LITERAL',
+            required: true,
+            description: 'encounter date/time in Fileman format',
+            example: '3170517.062721',
+        }],
+
+        invokeRPC: (rpcArgs, session, callback) => {
+            if (_.isEmpty(rpcArgs) || rpcArgs.length !== 2) return callback(new Error('Empty or incorrect length for rpcArgs'));
```
mfuroyama: You can delete this line. The built-in validation should take care of it.

emulation/models/orwpce-active-prov.js
```diff
@@ -0,0 +1,49 @@
...
+            const visitDate = rpcArgs[1] || '';
+            if (visitDate < 1800000) return callback(null, 0);
+            const provIen = rpcArgs[0] || '';
+            const provId = rpcArgs[0] ? modelUtils.createURNFromParams('200', provIen, session.stationId) : '';
```
mfuroyama: So remember the functional philosophy we're following here: we don't want to let users look at other users stuff. So let's do the same thing we've done in other RPCs here: check to see if the specified user is the same as what's in the context and fail out if it's not

services/pceMetaService/isActiveProvider.js
```diff
@@ -0,0 +1,49 @@
...
+ * @param  {String}   provId  the Provider ID to check
+ * @param  {String}   visitDate  encounter date/time in Fileman format
+ * @param  {Object}   context  Session context information associated with the service call.
+ * @param  {Function}   callback  Function called on process completion, with standard (err, result) signature.
+ * @return  {Boolean}  Either true or false depending on if provider is active or has an active person class
+ *
+ * @memberof PceMetaService
+ * @instance
+ */
+function isActiveProvider(provId, visitDate, context, callback) {
```
mfuroyama: So, let's remove the `provId` from the function signature (and the JSDoc comments too!) to prevent people from looking at other user's settings. The provider `id` should always come from the `context` object.

emulation/models/orwpce-active-prov.js
creindle: !!!! I should've caught that! Ahhhh... Ok, yes I'll fix this right now.

services/pceMetaService/isActiveProvider.js
creindle: Okok, this is good to know for sure!

services/pceMetaService/isActiveProvider.js
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * @summary This calls the PCE API $$ACTIVPRV^PXAPI(provider ien, encounter d/t) tosee if the provider can be stored by PCE.   Returns true if provider is good and false if the provider is not active or does not have an active person class.
+ * @param  {String}   provId  the Provider ID to check
+ * @param  {String}   visitDate  encounter date/time in Fileman format
+ * @param  {Object}   context  Session context information associated with the service call.
+ * @param  {Function}   callback  Function called on process completion, with standard (err, result) signature.
+ * @return  {Boolean}  Either true or false depending on if provider is active or has an active person class
+ *
+ * @memberof PceMetaService
```
mfuroyama: I think this is generic enough to where it can go into `UserMetaService`. Let's move this over

services/pceMetaService/isActiveProvider.js
```diff
@@ -6,43 +6,48 @@ const moment = require('moment');
...
-        if (terminationDate && terminationDate < visitDate) return false;
-        const activeClassList = res[0].personClass;
-        if (!activeClassList) return callback(null, false);
+        const terminationDate = _.get(res[0], 'terminationDate', '');
+        if (terminationDate && terminationDate < visitDate) return callback(null, false);
+        const activeClassList = _.get(res[0], 'personClass', []);
+        if (_.isEmpty(activeClassList)) return callback(null, false);
         const isValidClass = activeClassList.some((pClass) => {
             let effectiveDate = _.get(pClass, 'effectiveDate', '');
+            effectiveDate = moment(effectiveDate).format('YYYY-MM-DDTHH:mm:ssZ');
```
creindle: Formatted Date() object to string to be used in xsdToFMDateTime

services/pceMetaService/isActiveProvider.js
```diff
@@ -0,0 +1,55 @@
...
+            return true;
+        });
+        if (!isValidClass) return false;
+    }).then(() => {
+        service.find('PersonClass', {
+            query: { id: personClassIds, status: 'Active' },
+        }).then((newPersonRes) => {
+            const newPersonResStatus = _.get(newPersonRes[0], 'status', false);
+            return callback(null, newPersonResStatus === 'Active');
+        });
```
mfuroyama: Hmm...'m having reservations about this nested promise chain. You know what? On Monday, let's talk about possibly using the new utility that @sidjtd created to handle multiple queries and conditions. This service function might be a perfect fit for that.

PR #182, opened 2018-01-03
vam 375 orwpt dfltsrc

pcchanghawaii:
implement the ORWPT DFLTSRC RPC without the mongo layer as the default source only return C (clinic) or undefined


emulation/models/orwpt-dfltsrc.js
```diff
@@ -0,0 +1,30 @@
...
+    labels: ['META', 'USER-NOT-REQUIRED', 'HARDCODED'],
+    services: [],
+    inputParameters: [],
+    invokeRPC: (args, session, callback) => {
+        userMetaService.lookupPatientListSource(session, callback);
+    },
+
+    formatResult: (res) => {
+        return res;
+    },
```
sidjtd: You can remove formatResult block if you are just returning the results as is, Mike created this set up up where when you don't change (or dont have to) modify the output to resemble the native mumps-y output!

emulation/models/orwpt-dfltsrc.js
```diff
@@ -0,0 +1,30 @@
...
+    function: 'DFLTSRC^ORWPT',
+    domain: 'PATIENT',
+    description: 'Return user’s default patient list source.',
+    returnType: 'SINGLE VALUE',
+    rpcType: 'READ',
+    complexity: 'LOW',
+    files: [],
+    parameters: [],
+    labels: ['META', 'USER-NOT-REQUIRED', 'HARDCODED'],
+    services: [],
```
sidjtd: In services we are supposed to add a tag like 'userMetaService.lookupPatientListSource', but since I think this is a clinical RPC (you put the domain down as 'PATIENT') I'm wondering if userMetaService is the right location? Hopefully Mike or someone else can help share their thoughts on this so we know how to proceed....

emulation/models/tests/orwpt-dfltsrc-spec.js
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env node
```
creindle: Since you are removing the `formatResult` method in the model file (orwpt-dfltsrc.js), you can remove this file since you would be testing the `formatResult` method that wouldn't exist anymore.

services/userMetaService/lookupPatientListSource.js
```diff
@@ -0,0 +1,34 @@
...
+ *                  C=clinics, W=wards, P=Providers T=Team/Personal, S=Specialities
+ * @memberof userMetaService
+ * @instance
+ */
+
+function lookupPatientListSource(context, callback) {
+    try {
+        // only ALEXANDER,ROBERT returns C other users return undefined [] according to RPC running results
+        let res;
+        const userId = parseInt(context.userId.split(':')[4], 10); // urn:vista:6100:200:63
```
creindle: This method can actually be replaced with the handy serviceUtils method `getIdFromUrn` that takes in an URN. So this would be reformatted as ```const userId = serviceUtils.getIdFromUrn(context.userId);```

creindle:
I apologize for adding quite a few comments to your PRs! If you have ANY questions, please let me know and I can help you out! I can even walk you through some of them too! For me, it would be very difficult to work on these PRs remotely, so good job!!

services/userMetaService/lookupPatientListSource.js
```diff
@@ -0,0 +1,34 @@
...
+ * @summary Return user’s default patient list source.
+ * @param {Object} context session information associated with the service call
+ * @param  {Function} callback Function called on process completion of lookupPatientListSource.
+ * @return {string} return default patient list source (T, W, C, P, S)
+ *                  C=clinics, W=wards, P=Providers T=Team/Personal, S=Specialities
+ * @memberof userMetaService
+ * @instance
+ */
+
+function lookupPatientListSource(context, callback) {
```
creindle: Hey Pei! Actually, after further research on this, we actually have a easier, pure JS implementation of what this should be doing according to the MUMPs implementation. [Here](https://github.com/vistadataproject/nonClinicalRPCs/wiki/MUMPS-Patterns-Demystified#parameters) is a description of this translation.
In the MUMPs code for this `ORWPT DFLTSRC`, it has this line: 
```Javascript
S VAL=$$GET^XPAR("ALL^SRV.`"_SRV,"ORLP DEFAULT LIST SOURCE")
``` 
Which can utilize the ParameterService method:
```Javascript
ParameterService.withContext(session).get('ORLP DEFAULT LIST SOURCE', { entity: ['ALL', 'SRV'] })
                .then((res) => {
                    const resultValue = _.get(res[0], 'value', '');
                    callback(null, resultValue);
                }).catch(err => callback(err));
```

What this is doing is going into the MongoDB and retrieving the value from the collection `Parameters` based on the `sessionContext.json` information. This value is one of the letters from the default patient list source (T, W, C, P, S).

Another example of this is MUMPS ['GETCTXT^ORQQCN2'](http://code.osehra.org/dox/Routine_ORQQCN2_source.html) and pure JS implemented [PR](https://github.com/vistadataproject/VICSServer/pull/77)

emulation/models/orwpt-dfltsrc.js
pcchanghawaii: https://github.com/vistadataproject/vdm/tree/master/prototypes#clinical-domains    The domain is a bit tricky as the RPC lists default patient list source for a user. Each user may have his or her preferred way of patient source.  It's basically cross domain between user and patient, or it's utility.  However, the RPC doesn't look like it's belong to the utilityRPC folder and I couldn't find the User domain from the above valid clinic domain, so I put Patient.

PR #183, opened 2018-01-05
Vam 103 lookup service connected category

sidjtd:
If comments need to be more succinct or more should be added, let me know! The comments will be added to the wiki as well


services/pceMetaService/lookupServiceConnectedCategory.js
```diff
@@ -0,0 +1,122 @@
...
+    const service = this;
+
+    /* ---------  Async Query Functions  --------- */
+
+    // Queries are separated into individual query modules. This particular module separates the query from response processing (see below)
+    function hospitalLocationQuery(cb) {
+        service.find('HospitalLocation', {
+            query: { id: locationUrn, stopCodeNumber: { $exists: true } },
+            projection: serviceUtils.createProjection(['id', 'stopCodeNumber']),
+        }).then(res => cb(null, res));
```
mfuroyama: The `service.find` function also takes an optional `callback` parameter. So, consider replacing this:
```Javascript
       }).then(code => cb(null, code));
```
with this:
```Javascript
       }, cb);
```
This not only makes the code cleaner, but it also handles the error conditions for you as well. If you don't use the callback scheme, you'll have to add a `catch` handler to your promise handling to catch errors...

services/pceMetaService/lookupServiceConnectedCategory.js
```diff
@@ -0,0 +1,122 @@
...
+            ** In the second argument function, do a conditional check for error in order to retrieve value from a possible early break.
+            ** Use if/else to separate the final callbacks in order to prevent the callback being called more than once */
+
+    async.waterfall([
+        hospitalLocationQuery,
+        hospitalProcess,
+        clinicStopQuery,
+        dssQuery,
+    ], (err, res) => {
+        if (_.has(err, 'returnValue')) callback(null, err.returnValue);
```
mfuroyama: Here too: don't forget to add code to handle the case where `err` truly contains a generated error.

PR #186, opened 2018-01-09
Vam 15 orwu userinfo

kekuso:
pls check my Promise chaining for errors, thanks!


services/userMetaService/listUserInfo.js
```diff
@@ -0,0 +1,208 @@
...
+
+    function domainResCheck(res, cb) {
+        // add the domain property to userObject
+        _.assign(userObject, { domain: res ? _.get(res[0], 'label', '') : '' });
+        cb(null);
+    }
+
+    function stationNumbertimeSensitiveQuery(cb) {
+        service.find('StationNumbertimeSensitive', {
+            query: { localStationNumber: 6100 },
```
mfuroyama: Let's use the `context.stationId` here instead of hard-coding `6100`

services/userMetaService/listUserInfo.js
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    function stationNumberCheck(res, cb) {
+        // add the station number property to userObject
+        _.assign(userObject, { stationNumber: res ? _.get(res[0], 'stationNumber', '') : -1 });
+        cb(null);
+    }
+
+    function kernelSystemParametersQuery(cb) {
+        service.find('KernelSystemParameters', {
```
mfuroyama: This is okay, but consider using the `systemMetaService/lookupKernelSystemParameters` service function here. The contents of this function would reduce down to this:
```Javascript
    systemService.lookupKernelSystemParameters(context, cb);
```

The system meta service function version also has the added benefit of some data caching built in: https://github.com/vistadataproject/VICSServer/blob/master/services/systemMetaService/lookupKernelSystemParameters.js

PR #189, opened 2018-01-10
VAM-479: Update `lookupSessionContext` service function 

mfuroyama:
The service function now takes into account user divisions, and stores the division information in the session object returned.


services/tests/lookupSessionContext-spec.js
```diff
@@ -7,43 +7,65 @@ const serviceTestUtils = require('./serviceTestUtils');
 // Each test case represents a different system contained within the MongoDB
 const TEST_CASES = [{
     name: 'nodevista',
-    userIEN: 63,
+    userId: 'urn:vista:6100:200:63',
```
creindle: Good catch!!

mfuroyama: I'm going to take your comment as an implicit pull request approval...

PR #190, opened 2018-01-10
VAM-39: ORQQPL USER PROB CATS

creindle:
[X] Imported Cheyenne data for problemSelectionList collection
[X] Replaced user IEN with session userId in capture files
[X] Passes comparison, model, and services tests


emulation/models/orqqpl-user-prob-cats.js
```diff
@@ -0,0 +1,48 @@
...
+
+const orderMetaService = require('../../services/orderMetaService');
+const modelUtils = require('./modelUtils');
+
+const RPC_EMULATOR_MODEL = [
+    {
+        id: 'ORQQPL_USER_PROB_CATS',
+        name: 'ORQQPL USER PROB CATS',
+        function: 'CAT^ORQQPL3',
+        domain: 'ORDER',
```
mfuroyama: I think this is part of the 'PROBLEMS' domain. Let's change this out

creindle: Oh! That makes sense because the label had 'PROBLEMS' on it!

emulation/models/orqqpl-user-prob-cats.js
```diff
@@ -0,0 +1,48 @@
...
+            parameterType: 'LITEARL',
+            required: true,
+            description: 'Clinical Location IEN',
+            example: '3',
+        }],
+        invokeRPC: (rpcArgs, session, callback) => {
+            const user = rpcArgs[0] ? modelUtils.createURNFromParams(200, rpcArgs[0], session.stationId) : '';
+            if (user !== session.userId) return callback(new Error('User IEN from Client does not match with session user IEN'));
+            const location = modelUtils.createLocationEntity(rpcArgs[1], session);
+            orderMetaService.lookupCategoriesForUser(location, session, callback);
```
mfuroyama: Ahh...unfortunately, this will need to be changed to `problemMetaService` as well, which means we'll need to create a new meta service... 😞  ...sorry...thanks!

services/orderMetaService/lookupCategoriesForUser.js
```diff
@@ -0,0 +1,64 @@
...
+ * @instance
+ */
+function lookupCategoriesForUser(location, context, callback) {
+    const service = this;
+    let userList;
+    let problemSelectionListContentsId;
+    const arrOfCategories = [];
+    function newPersonQuery(cb) {
+        service.find('NewPerson', {
+            query: { id: context.userId },
```
mfuroyama: Consider adding a projection to your query that only returns `problemSelectionList` since it looks like that's all you're using from the user (NewPerson) object coming back

emulation/models/orqqpl-user-prob-cats.js
creindle: That's fine! Thank you for the extra check!

PR #192, opened 2018-01-16
VAM 410: ORWPCE I10IMPDT

mfuroyama:
One of two Lexicon Service RPCs. This is the easier of the two and sets up the Lexicon Service infrastructure.


mfuroyama:
@conorcaregraf LOL! I just realized I auto-completed the incorrect function name everywhere in the code!

Anyway, good point about the function! The `loookupSourceImplementationDate` function gets the implementation date of a specified coding system (757.03), but yeah, it will probably never be used. I think I'll change this to 'lookup' a Coding System, then do the extraction in the RPC model.

PR #193, opened 2018-01-16
VAM-38: ORQQPL INIT USER

creindle:


services/problemMetaService/getUserParametersForProblemList.js
```diff
@@ -0,0 +1,109 @@
...
+    }
+    function gmplProvNewPersonQuery(gmplProvId, cb) {
+        service.find('NewPerson', {
+            query: { id: gmplProvId },
+        }, cb);
+    }
+    function gmplProvNewPersonProcess(res, cb) {
+        const provUser = !_.isEmpty(res) ? res[0] : '';
+        response[7] = provUser ? `${gmplProv}^${provUser.name}` : '0^All';
+        response[8] = '';
```
creindle: nonClinicalRPC version of this calls for function `getService(userId)` which returns the service section name in a mumps format and sets it as this value. If we need this function, let me know and I can add it in.

emulation/models/orqqpl-init-user.js
```diff
@@ -0,0 +1,43 @@
...
+        id: 'ORQQPL_INIT_USER',
+        name: 'ORQQPL INIT USER',
+        functloion: 'INITUSER^ORQQPL1',
+        domain: 'PROBLEM',
+        parameters: ['ORCH CONTEXT PROBLEMS', 'ORQQPL SUPPRESS CODES'],
+        files: ['49', '200'],
+        labels: ['META'],
+        services: ['problemMetaService.getUserParametersForProblemList'],
+        rpcType: 'READ',
+        complexity: 'MODERATE/DIFFICULT',
```
mfuroyama: Let's just pick one here...I'd suggest 'DIFFICULT'

services/problemMetaService/getUserParametersForProblemList.js
mfuroyama: What?

services/problemMetaService/getUserParametersForProblemList.js
```diff
@@ -0,0 +1,109 @@
...
+const serviceUtils = require('../serviceUtils');
+const userMetaService = require('../userMetaService');
+const ParameterService = require('../parameterService');
+const _ = require('lodash');
+
+/**
+ * @summary Returns user parameters for problem list
+ * @param  {Object}   context  Session context information associated with the service call.
+ * @param  {Function}   callback  Function called on process completion, with standard (err, result) signature.
+ * @return {Array}   response  User parameters for problem list
```
mfuroyama: You might want to consider documenting exactly what the array values correspond to

services/problemMetaService/getUserParametersForProblemList.js
```diff
@@ -0,0 +1,109 @@
...
+    function newPersonProcess(res, cb) {
+        const probListPrimaryView = _.get(res[0], 'probListPrimaryView', '');
+        const viewItems = !_.isEmpty(probListPrimaryView) ? probListPrimaryView.split('/') : '';
+        const item = !_.isEmpty(viewItems) ? viewItems[0] : '';
+        const gmplView = {};
+
+        if (viewItems.length > 1) {
+            if (item === 'C') gmplView.clin = viewItems.slice(1, 100);
+            if (item === 'S') gmplView.serv = viewItems.slice(1, 100);
+        }
```
mfuroyama: Could you explain to me exactly what's going on here?

services/problemMetaService/getUserParametersForProblemList.js
```diff
@@ -0,0 +1,109 @@
...
+        cb(null, null);
+    }
+    function orchContextProblemsParamQuery(res, cb) {
+        ParameterService.withContext(context).get('ORCH CONTEXT PROBLEMS', { entity: 'ALL' }, cb);
+    }
+    function orchContextProblemsParamProcess(res, cb) {
+        const contextParam = _.get(res[0], 'value', '');
+        gmplProv = !_.isEmpty(contextParam) ? contextParam.split(';')[4] : 0;
+        response[6] = contextParam.split(';')[2] !== '' ? contextParam.split(';')[2] : 'A';
+        response[12] = parseInt(contextParam.split(';')[3], 10) || 0;
```
mfuroyama: Here, instead of repeatedly performing a `contextParam.split()` operation, consider doing the split on `contextParam` value on `;` just once and re-using the results in all these comparisons.  Doing this would make the code slightly more efficient *and* readable.

services/problemMetaService/getUserParametersForProblemList.js
```diff
@@ -0,0 +1,109 @@
...
+        cb(null, null);
+    }
+    function problemListSiteParamQuery(res, cb) {
+        service.find('ProblemListSiteParameters', {
+            query: { localStationNumber: context.stationId },
+        }, cb);
+    }
+    function problemListSiteParamProcess(res, cb) {
+        response[2] = _.get(res[0], 'verifyTranscribedProblems', false);
```
mfuroyama: Consider assigning the value of `res[0]` here to a variable, to improve readability.

services/problemMetaService/getUserParametersForProblemList.js
```diff
@@ -0,0 +1,109 @@
...
+        const gmplView = {};
+
+        if (viewItems.length > 1) {
+            if (item === 'C') gmplView.clin = viewItems.slice(1, 100);
+            if (item === 'S') gmplView.serv = viewItems.slice(1, 100);
+        }
+        response[1] = !_.isEmpty(viewItems) ? item : probListPrimaryView;
+        response[9] = gmplView.serv || '';
+        response[10] = gmplView.clin || '';
+        cb(null, null);
```
mfuroyama: Here (and in other places) where you **don't** need to pass results to the next function, you can just call `cb(null);`, then in the subsequent function in the list, just make sure that the function signature only takes `cb` as an argument...so in this particular example, the `orchContextProblemsParamQuery` function would look like this:
```Javascript
 function orchContextProblemsParamQuery(cb) {
```

services/problemMetaService/getUserParametersForProblemList.js
creindle: That's a good idea!

services/problemMetaService/getUserParametersForProblemList.js
creindle: I wasn't too sure what this was doing as I couldn't find any data to format this information, so I copied this over from the nonClinicalRPC implementation. The nonClinicalRPC has gmplView as an empty object too after these lines of code.

services/problemMetaService/getUserParametersForProblemList.js
creindle: Thank you for the clarification! I have to double check if my previous implementations utilize this.

services/problemMetaService/getUserParametersForProblemList.js
creindle: For future notes, in nonClinicalRPC, this line was `res[8] = getService(userId);` and the method was ```function getService(id, incnpc) {
                let x = user.serviceSection.id.split('-')[1];
                const serviceSection = emulatorUtils.tryMVDMDescribe(`49-${x}`);
                if (serviceSection.error) return '';
                const _incnpc = parseInt(incnpc, 10) || 0;
                if (!_incnpc && serviceSection.name.split('^')[8] !== 'C') x = 0;
                return (x > 0) ? `${x}^${serviceSection.name}` : '';
            }```

services/problemMetaService/getUserParametersForProblemList.js
creindle: After realizing that I was retrieving the data for this incorrectly, I believe this is setting the object gmplView keys to viewItems[1] depending on what item was. Initially, the data coming in from problemListPrimaryView came in as `C/`, but it seems from the nonClinicalRPC version of this, that some values had longer strings, possibly like `C/ExtraWord`. Not 100% sure because the database didn't have a case where the length of viewItems would be more than 1.

services/userMetaService/hasMultipleSecurityKeys.js
```diff
@@ -0,0 +1,52 @@
+'use strict';
+
+const serviceUtils = require('../serviceUtils');
+const _ = require('lodash');
+
+function checkMultipleSecurityKeysWithNewPersonRes(key, newPersonRes, cb) {
```
mfuroyama: Couple things:
1) Consider shortening the name of this function to something like `checkUserSecurityKeys`
2) Consider adding JSDoc comments to this function to document the API, since you're publishing this function as an API by including it in the module exports.

services/userMetaService/hasMultipleSecurityKeys.js
```diff
@@ -0,0 +1,52 @@
...
+    const hasKey = (_.isArray(key) && !_.isEmpty(key)) ?
+        key.map(k => _.includes(labels, k)) :
+        (labels && _.includes(labels, key));
+    if (!_.includes(hasKey, true) || !hasKey) return cb(null, false);
+    return cb(null, true);
+}
+
+/**
+ * @summary Returns true if a user holds a security key, otherwise false.
+ * @param  {Array}   key      The names of the security keys (i.e. ['GMV MANAGER', 'LRLAB'])
```
mfuroyama: Minor readability improvement suggestion: consider renaming the key parameter to keys. This conveys implied intent to the user that that parameter, is, in fact, a list of key values.

services/userMetaService/hasMultipleSecurityKeys.js
```diff
@@ -0,0 +1,52 @@
...
+    function newPersonProcess(res, cb) {
+        checkMultipleSecurityKeysWithNewPersonRes(key, res, cb);
+    }
+    function returnResponse(res, cb) {
+        cb(null, res);
+    }
+    const allFunctions = [
+        newPersonQuery,
+        newPersonProcess,
+        returnResponse,
```
mfuroyama: I'm pretty sure this function serves no practical purpose. Let's remove it from the list

services/userMetaService/hasMultipleSecurityKeys.js
creindle: This makes sense! Thank you so much for your suggestions and writing a checklist!

services/userMetaService/hasMultipleSecurityKeys.js
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * @summary Returns true if the provided user holds a set of security keys, otherwise false.
+ * @param  {Array}   keys  The names of the security keys (i.e. ['GMV MANAGER', 'LRLAB'])
+ * @param  {Object}   newPersonRes NEW PERSON (200) query response
+ * @return {Boolean}
+ */
+function checkUserSecurityKeys(keys, newPersonRes) {
+    if (!newPersonRes || !newPersonRes.length) return false;
+    const resKeys = _.get(newPersonRes[0], 'keys', '');
```
mfuroyama: A better default value to return from the `_.get` would be an empty array.

services/userMetaService/hasMultipleSecurityKeys.js
```diff
@@ -0,0 +1,54 @@
...
+    const labels = resKeys ? resKeys.map(k => _.get(k, 'key.label', '')) : '';
```
mfuroyama: Here's a nifty little trick you can do with lodash if you want to extract fields out of an object. You can replace the above line with:
```Javascript
const labels = _.map(resKeys, 'key.label');
```

Try it out! It relies on the fact that lodash `map` applies `_.property` to string arguments.

services/userMetaService/hasMultipleSecurityKeys.js
```diff
@@ -0,0 +1,54 @@
...
+    const hasKeys = (_.isArray(keys) && !_.isEmpty(keys)) ?
```
mfuroyama: Consider moving the 'empty array' check to the top of the function. Better to bail early if error conditions exist.

services/userMetaService/hasMultipleSecurityKeys.js
```diff
@@ -0,0 +1,54 @@
...
+        keys.map(k => _.includes(labels, k)) :
+        (labels && _.includes(labels, keys));
+    if (!_.includes(hasKeys, true) || !hasKeys) return false;
+    return true;
```
mfuroyama: Consider replacing the above block of logic (which is slightly hard to follow) with something slightly simpler. One method to consider would be to use the lodash `intersection` function. Compare the 'keys' input array with the user key labels. If all the keys are present, true, otherwise, false:
```Javascript
return _.intersection(keys, labels).length === keys.length;
```

creindle: That's really nifty!

services/userMetaService/hasMultipleSecurityKeys.js
```diff
@@ -0,0 +1,50 @@
+'use strict';
+
+const serviceUtils = require('../serviceUtils');
+const _ = require('lodash');
+
+/**
+ * @summary Returns true if the provided user holds a set of security keys, otherwise false.
+ * @param  {Array}   keys  The names of the security keys (i.e. ['GMV MANAGER', 'LRLAB'])
+ * @param  {Array}   newPersonRes NEW PERSON (200) query response containing keys
```
mfuroyama: Shoot...sorry, I just noticed this: Any particular reason why this is an Array of users (instead of a single user object)? I think it might be better to adjust the function to accept a single user (NewPerson) object and adjust your calling functionality accordingly.

creindle: There's no particular reason! I usually pass in the entire response from a query into a function, but for a single user (NewPerson), you're right, I should pass in the single person object!

PR #199, opened 2018-01-19
VAM-137: TIU IS USER A PROVIDER?

creindle:
[X] Passes comparison, model, and service tests


services/userMetaService/getActiveClass.js
```diff
@@ -0,0 +1,88 @@
...
+    const service = this;
+    const userActiveClass = {};
+    function newPersonQuery(cb) {
+        service.find('NewPerson', {
+            query: { id: context.userId },
+            projection: serviceUtils.createProjection(['personClass']),
+        }, cb);
+    }
+    function newPersonProcess(res, cb) {
+        if (!_.isArray(res) || _.isEmpty(res)) return false;
```
mfuroyama: Remember that when terminating early from within an **asynchronous** function, the right way to go is to call the `cb` function with the first (error) parameter set.  Merely returning false will cause the application to hang because the asynchronous execution path is relying on all the callback functions to be called. So here:
```Javascript
if (!_.isArray(res) || _.isEmpty(res)) {
    return cb(new Error('Some error message'));
}
```

A good way to test this (and a "best practice" for testing as well) is to add some "negative" test cases in your Jasmine service test specs. Set up a spec or two that will cause `res` to be empty and cause the function to fail at this point. (hint: try passing in a context object with an invalid `userId`)

Remember that the point of writing Jasmine test cases is to ensure that you cover (and execute) as many possible execution branches and paths of your code as possible.

services/userMetaService/getActiveClass.js
```diff
@@ -0,0 +1,88 @@
...
+        const activeClassList = _.get(res[0], 'personClass', []);
+        let activeClassEffectiveDate = '';
+        const personClassExpirationDateList = [];
+        if (!activeClassList) return false;
+        if (parseInt(visitDate, 10) < 1800000 || !visitDate.toString().match(/\d{7}\.\d{6}|\d{7}/)) return false;
+        _.each(activeClassList, (pClass) => {
```
mfuroyama: Several things to consider to improve readability:
1) Instead of mere looping with an `_.each`, consider breaking this up into a `filter` and a `map`. So, `filter` the person class array against the visit date, then `map` the remaining collection on `id`. I think this would make the code a little bit more readable, since `filter` and `map` imply what the code is actually doing.
2) Do the `parseInt(visitDate.split('.')[0], 10))` once before the looping and use the value in the loop to improve readability and efficiency.
3) I'm trying to figure out why you declared `activeClassEffectiveDate` outside the looping `each`. It adds to the cognitive load and you probably don't need it out there.

services/userMetaService/getActiveClass.js
```diff
@@ -0,0 +1,88 @@
...
+        }, cb);
+    }
+    function checkExpirationDate(res, cb) {
+        const personClassObj = res;
+        let expirationDateFm = '';
+        const isValidClass = _.some(personClassObj, (pClass) => {
+            const expirationDate = _.get(pClass, 'expirationDate', '');
+            if (expirationDate) {
+                expirationDateFm = dateTimeService.xsdToFMDateTime(expirationDate.toISOString());
+                if (parseInt(expirationDateFm, 10) < parseInt(visitDate.split('.')[0], 10)) return false;
```
mfuroyama: Same comment here: since you're using this parsed, numeric version of the FM date in so many places, consider just doing it once at the top-most level of this function and use that value instead of repeatedly split/parsing `visitDate`

services/userMetaService/getActiveClass.js
```diff
@@ -0,0 +1,88 @@
...
+        const returnPersonClass = {
+            personClassId: serviceUtils.getIdFromUrn(personClassObjId),
+            occupation: personClassObjOccupation,
+            speciality: personClassObjSpeciality,
+            subSpeciality: personClassObjSubSpeciality,
+            effectiveDate: userActiveClass.activeClassEffectiveDate,
+            expirationDate: expirationDateFm,
+            vaCode: personClassObjVaCode,
+            specialtyCode: personClassObjSpecialtyCode,
+
```
mfuroyama: Just a matter of style, but consider possibly cleaning this up by either:

Giving the transitory variables names that match the final object keys. For example, instead of:
```Javascript
const personClassObjOccupation = _.get(personClassObj, 'providerType', '');
```

Do this instead:

```Javascript
const occupation = _.get(personClassObj, 'providerType', '');
```
Then, when you declare your object literal, you can take advantage of the ES6 naming shortcut:
```Javascript
        const returnPersonClass = {
            ...
            occupation: _.get(personClassObj, 'providerType', ''),
            ...
        };
```


Either that, or just forego the temporary variable altogether:
```Javascript
        const returnPersonClass = {
            ...
            occupation,
            ...
        };
```

services/userMetaService/isValidProvider.js
```diff
@@ -0,0 +1,81 @@
...
+        const usrClassMembershipId = serviceUtils.createURNFromParams('8930_3', userId, userObj.localStationNumber);
+        service.find('UsrClassMembership', {
+            query: { id: usrClassMembershipId },
+            projection: serviceUtils.createProjection(['userClass']),
+        }, cb);
+    }
+    function usrClassQuery(res, cb) {
+        let usrClassId = _.get(res[0], 'userClass.id', '');
+        usrClassId = serviceUtils.getIdFromUrn(usrClassId);
+        const usrClassQueryId = serviceUtils.createURNFromParams(8930, usrClassId, userObj.localStationNumber);
```
mfuroyama: Correct me if I'm wrong here, but it looks like what you're doing here is extracting an IEN out of a URN, then rebuilding the URN with the IEN. I'm pretty sure you don't need to do this.

You might also want to add a check if the previous query returns no results

services/userMetaService/isValidProvider.js
```diff
@@ -0,0 +1,81 @@
...
+        service.find('UsrClass', {
+            query: { id: usrClassQueryId },
+            projection: serviceUtils.createProjection(['active', 'label']),
+        }, cb);
+    }
+    function usrClassProcess(res, cb) {
+        const usrClassActive = _.get(res[0], 'active', '');
+        const usrClassName = _.get(res[0], 'name', '');
+        isProvider = usrClassActive === 'Inactive' ? false : usrClassName === 'PROVIDER';
```
mfuroyama: Minor readability improvement suggestion: logically speaking, `isProvider` is true if:
* `usrClassActive` is not 'Inactive' and
* `usrClassName` is 'PROVIDER'
correct? If so, then it might be clearer to write this as:
```Javascript
isProvider = ((usrClassActive !== 'Inactive') && (usrClassName === 'PROVIDER'));
```
It's subtle, but see how it conveys intent a little more clearly in this case?

services/userMetaService/isValidProvider.js
```diff
@@ -0,0 +1,81 @@
...
+    }
+    function personClassProcess(res, cb) {
+        const personClassEffectiveDate = _.get(res[0], 'effectiveDate', '');
+        const personClassEffectiveDateFm = personClassEffectiveDate ? dateTimeService.xsdToFMDateTime(personClassEffectiveDate) : '';
+        if (parseInt(personClassEffectiveDateFm, 10) > parseInt(visitDate, 10)) cb(null, false);
+        const status = _.get(res[0], 'status', '');
+        const vaCode = _.get(res[0], 'vaCode', '');
+        if (status === 'Active'
+            && vaCode
+            && !vaCode.match(/(^V0802|V0805|V0806|V0808|V0809|V0812|V0813).*/)) cb(null, true);
```
mfuroyama: I think you were intending to write:
```Javascript
return cb(null, true);
```

services/userMetaService/isValidProvider.js
```diff
@@ -0,0 +1,81 @@
...
+        const personClassId = serviceUtils.createURNFromParams('8932_1', userId, userObj.stationId);
+        service.find('PersonClass', {
+            query: { id: personClassId },
+            projection: serviceUtils.createProjection(['status', 'vaCode', 'effectiveDate']),
+        }, cb);
+    }
+    function personClassProcess(res, cb) {
+        const personClassEffectiveDate = _.get(res[0], 'effectiveDate', '');
+        const personClassEffectiveDateFm = personClassEffectiveDate ? dateTimeService.xsdToFMDateTime(personClassEffectiveDate) : '';
+        if (parseInt(personClassEffectiveDateFm, 10) > parseInt(visitDate, 10)) cb(null, false);
```
mfuroyama: Same here:
```
return cb(null, false);
```

services/userMetaService/getActiveClass.js
creindle: Thank you for that catch! I'm still getting used to these asynchronous callbacks!

services/userMetaService/getActiveClass.js
creindle: Thank you for the thorough explanation!

services/userMetaService/isValidProvider.js
creindle: Ah! You're right! I was so used constructing the ids, I didn't realize that the filetype matched from the extracted id for the second query! Thank you for that catch!

services/userMetaService/isValidProvider.js
creindle: This looks a lot more readable!

services/userMetaService/isValidProvider.js
creindle: It sucks I missed all of these for this RPC!

creindle: And if it doesn't `return` then it would continue with the code!

services/documentMetaService/isUserAProvider.js
```diff
@@ -0,0 +1,50 @@
...
+ * @param  {Object}   context  Session context information associated with the service call.
+ * @param  {Function} callback Function called on process completion, with standard (err, result) signature.
+ * @return {Boolean}  returns true if the user was a known provider on the date specified
+ *
+ * @memberof DocumentMetaService
+ * @instance
+ */
+function isUserAProvider(visitDate, context, callback) {
+    const service = this;
+    let userObj = {};
```
mfuroyama: Any particular reason you're declaring `userObj` out here using `let`? Consider removing this declaration here, and using `const userObj` declaration in `isValidProviderCall`

services/documentMetaService/isUserAProvider.js
```diff
@@ -0,0 +1,50 @@
...
+        userMetaService.isValidProvider(userObj, visitDate, cb);
+    }
+    function getActiveClassCall(res, cb) {
+        isValidProvider = res;
+        userMetaService.getActiveClass(visitDate, context, cb);
+    }
+    function isUserAProviderProcess(res, cb) {
+        getActiveClass = res;
+        if (isValidProvider || getActiveClass) return cb(serviceUtils.bail(true));
+        cb(null, false);
```
mfuroyama: I'm guessing this is probably a product of iterative editing, but I believe you can replace the 2 lines above with:
```Javascript
cb(null, (isValidProvider || getActiveClass));
```

services/userMetaService/getActiveClass.js
```diff
@@ -0,0 +1,83 @@
...
+            query: { id: context.userId },
+            projection: serviceUtils.createProjection(['personClass']),
+        }, cb);
+    }
+    function newPersonProcess(res, cb) {
+        if (!_.isArray(res) || _.isEmpty(res)) return cb(serviceUtils.bail(false));
+        const activeClassList = _.get(res[0], 'personClass', []);
+        if (!activeClassList) return cb(serviceUtils.bail(false));
+        if (parseInt(visitDate, 10) < 1800000 || !visitDate.toString().match(/\d{7}\.\d{6}|\d{7}/)) return cb(serviceUtils.bail(false));
+        const personClassId = _.chain(activeClassList)
```
mfuroyama: Lodash has a nifty little shortcut to set up a `chain` operation. Instead of:
```Javascript
const personClassId = _.chain(activeClassList)
```

You can just write:
```Javascript
const personClassId = _(activeClassList)
```

services/userMetaService/getActiveClass.js
```diff
@@ -0,0 +1,83 @@
...
+            .filter((pClass) => {
+                let activeClassEffectiveDate = _.get(pClass, 'effectiveDate', '');
+                if (!activeClassEffectiveDate) return cb(serviceUtils.bail(false));
```
mfuroyama: In the `filter` iteratee function, you should only be returning `true` or `false`. Calling the callback:
```Javascript
return cb(serviceUtils.bail(false))
```
Will produce undesirable side-effects (like potentially calling a called callback...say that 10 times quick)

Let's change this here:
```Javascript
return cb(serviceUtils.bail(false));
```

To just this:
```Javascript
return false;
```

services/userMetaService/getActiveClass.js
```diff
@@ -0,0 +1,83 @@
...
+        const returnPersonClass = {
+            id: serviceUtils.getIdFromUrn(id),
+            occupation: _.get(personClassObj, 'providerType', ''),
+            speciality: _.get(personClassObj, 'classification', ''),
+            subSpeciality: _.get(personClassObj, 'areaOfSpecialization', ''),
+            effectiveDate: userActiveClass.activeClassEffectiveDate,
+            expirationDate: expirationDateFm,
+            vaCode: _.get(personClassObj, 'vaCode', ''),
+            specialityCode: _.get(personClassObj, 'specialtyCode', ''),
+
```
mfuroyama: Really super, ultra minor, but let's get rid of this extra line...which for some reason wasn't flagged in ESLint but I think should have been.

services/documentMetaService/isUserAProvider.js
creindle: That's a cool way to condense this!

services/documentMetaService/isUserAProvider.js
creindle: There was no particular reason why it was created outside. I may've had it in my original implementation before I broke up the code in separate functions.

services/userMetaService/getActiveClass.js
creindle: Ooooooooh. Thank you for that catch!

services/userMetaService/getActiveClass.js
creindle: That's completely fine! I don't know why it wasn't flagged either!!

services/userMetaService/getActiveClass.js
creindle: That's so cool!

services/userMetaService/getActiveClass.js
```diff
@@ -0,0 +1,82 @@
...
+        const personClassId = _(activeClassList)
+            .filter((pClass) => {
+                let activeClassEffectiveDate = _.get(pClass, 'effectiveDate', '');
+                if (!activeClassEffectiveDate) return false;
+                activeClassEffectiveDate = activeClassEffectiveDate.toISOString();
+                activeClassEffectiveDate = dateTimeService.xsdToFMDateTime(activeClassEffectiveDate);
+                return activeClassEffectiveDate < parsedVisitDate;
+            })
+            .map(pClass => pClass.personClass.id)
+            .value();
```
mfuroyama: Yeah, this is **way** cleaner! 👍 

See...from here, I can easily tell at a glance what the code is meant to do based on if it's in either the `filter` block or the `map` block! Good stuff!

creindle: Thank you!

PR #200, opened 2018-01-22
VAM-73: ORWDXM FORMID

creindle:
[X] Passes comparison, model, service, and model validity tests


emulation/models/orwdxm-formid.js
```diff
@@ -0,0 +1,36 @@
...
+    services: ['orderMetaService.lookupDialogFormId'],
+    inputParameters: [{
+        id: 'DLG',
+        parameterType: 'NUMBER',
+        required: true,
+        description: 'the dialog id number',
+        example: '140',
+    }],
+    invokeRPC: (rpcArgs, session, callback) => {
+        if (!rpcArgs[0] || !_.isNumber(+rpcArgs[0])) return callback(new Error('Incorrect argument, argument is not defined or is not Number'));
```
mfuroyama: The internal validator will take care of ensuring the presence and type of the args based on the `inputParameters` objects, so this code becomes redundant. Let's remove this from here.

services/orderMetaService/lookupDialogFormId.js
```diff
@@ -0,0 +1,89 @@
...
+ */
+function lookupDialogFormId(dialogId, context, callback) {
+    const service = this;
+    let orderDialogObj = {};
+    let defDialogId = '';
+    let packageId = '';
+    let displayGroup = '';
+    let formId = false;
+
+    const orderDialogId = serviceUtils.createURNFromParams('101_41', dialogId, context.stationId);
```
mfuroyama: Let's move this conversion to URN to the RPC model `invokeRPC` implementation. `dialogId` is actually an IEN and we want the service functions to only deal with URNs.

Of course, this implies that the `dialogId` parameter in the JSDoc comments need to be updated as well.

services/orderMetaService/lookupDialogFormId.js
```diff
@@ -0,0 +1,89 @@
...
+        orderDialogObj = res[0];
+        if (!orderDialogObj || orderDialogObj.error) return cb(serviceUtils.bail(false));
+
+        formId = _.get(orderDialogObj, 'windowFormId', '');
+        if (_.isNumber(formId)) return cb(serviceUtils.bail(formId));
+
+        const orderDialogType4 = _.get(orderDialogObj, 'type4', '');
+        let orderDialogDisplayGroupId = _.get(orderDialogObj, 'displayGroup.id', '');
+        orderDialogDisplayGroupId = serviceUtils.getIdFromUrn(orderDialogDisplayGroupId);
+        displayGroup = (orderDialogType4 === 'quick order') ? orderDialogDisplayGroupId : '';
```
mfuroyama: I'm trying to figure out the intent from this segment of code here, and I'm having a hard time. A couple things I notice:
1) You're converting `orderDialogDisplayGroupId` to an IEN and passing that as the `id` query criteria in the `DisplayGroup` collection. All the `id` fields in the collections are URNs, so I don't think this will ever succeed. I *think* you meant to pass the URN in?
2) What should happen if `orderDialogType4` is not equal to `'quick order'`? Here, you're passing an empty string into the query, but I don't think this is what you want to do?

services/orderMetaService/lookupDialogFormId.js
```diff
@@ -0,0 +1,89 @@
...
+    function getDefaultDialog(res, cb) {
+        const displayGroupObj = res[0];
+        defDialogId = _.get(displayGroupObj, 'defaultDialog', '');
+
+        formId = (_.isNumber(displayGroup)) ? defDialogId : formId;
+        if (_.isNumber(formId)) return cb(serviceUtils.bail(formId));
+
+        service.find('DisplayGroup', {
+            query: { id: displayGroup },
+        }, cb);
```
mfuroyama: I'm so confused here...it seems like you're performing the exact same query here in consecutive functions. Was this your intent, or am I misunderstanding something?

services/orderMetaService/lookupDialogFormId.js
```diff
@@ -0,0 +1,89 @@
...
+    }
+    function getDisplayGroup(res, cb) {
+        const displayGroupObj = res[0];
+        let displayGroupId = _.get(displayGroupObj, 'id', '');
+        displayGroupId = serviceUtils.getIdFromUrn(displayGroupId);
```
mfuroyama: Again, not sure why this is being converted to an IEN here. Remember that service functions deal in URNs

services/orderMetaService/lookupDialogFormId.js
```diff
@@ -0,0 +1,89 @@
...
+            if (_.isNumber(formId)) return cb(serviceUtils.bail(formId));
+        }
+        packageId = _.get(orderDialogObj, 'package.id', '');
+        service.find('Package', {
+            query: { id: packageId },
+        }, cb);
+    }
+    function checkPackageId(res, cb) {
+        const crossReferencePackageId = _.get(res[0], 'id', '');
+        formId = packageId === crossReferencePackageId ? 152 : formId;
```
mfuroyama: What's the deal with the mystery, hard-coded value of '152' here?

mfuroyama:
@creindle Let's talk about the code here. I'm having a little bit of difficulty understanding the overall intent, and I want you to help me out here before I proceed with the code review.

services/orderMetaService/lookupDialogFormId.js
creindle: I am not sure what this value is used for! I saw it in the nonClinicalRPC version of this RPC

services/orderMetaService/lookupDialogFormId.js
creindle: Yes, this was definitely a confusing RPC to me too. I believe the issue may be that I'm misinterpreting this line incorrectly `const getIenDefaultDialogAd = ien => eUtils.lookupCrossReference('100.98', 'ORD', 'AD', [ien]);`

services/orderMetaService/lookupDialogFormId.js
creindle: Yes! You're correct, I meant to pass in a URN.

creindle: Also, it's weird for your second question, I saw that an empty string was passed in in the nonClinicalRPC implementation... So I would love some clarification here too!

services/orderMetaService/lookupDialogFormId.js
```diff
@@ -0,0 +1,84 @@
...
+        formId = _.get(orderDialogObj, 'windowFormId', '');
+        if (formId) return cb(serviceUtils.bail(formId));
+
+        const orderDialogType4 = _.get(orderDialogObj, 'type4', '');
+        const orderDialogDisplayGroupId = _.get(orderDialogObj, 'displayGroup.id', '');
+        // If order type is quick order, get display group type
+        displayGroup = (orderDialogType4 === 'quick order') ? orderDialogDisplayGroupId : '';
+        service.find('DisplayGroup', {
+            query: { id: displayGroup },
+        }, cb);
```
mfuroyama: Actually, I think a more accurate representation of the original nonClinicalRPC code here would be to only perform the query if `displayGroup` is valid:
```Javascript
if (displayGroup) {
    service.find('DisplayGroup', {
        query: { id: displayGroup },
    }, cb);
} else {
    cb(null, null);
}
```

This way, you don't make any unnecessary queries

services/orderMetaService/lookupDialogFormId.js
```diff
@@ -0,0 +1,84 @@
...
+    }
+    function getDisplayGroup(res, cb) {
+        const displayGroupObj = res[0];
```
mfuroyama: According to the original code, if `displayGroupObj.defaultDialog` is valid, you can bail here:
```Javascript
if (_.has(displayGroupObj, 'defaultDialog')) {
    cb(serviceUtils.bail(displayGroupObj.defaultDialog));
}
```

services/orderMetaService/lookupDialogFormId.js
creindle: That makes a lot more sense!

services/orderMetaService/lookupDialogFormId.js
mfuroyama: If the query result is invalid, then you'll need to query the DisplayGroup collection to find entries whose `member.member.id` matches `displayGroup`, then check those entries for valid `defaultDialog.id`:
```Javascript
service.find('DisplayGroup', {
    query: { 'member.member.id': displayGroup },
}, cb);
```

If you find a valid entry, bail on that, otherwise do the Package query and return the results...pretty simple!

mfuroyama: Also of note. This:
```Javascript
const crossReferencePackageId = eUtils.lookupCrossReference('9.4', 'DIC', 'C', ['OR']);
```

Is the same as this:
service.find('Package', {
    query: { prefix: 'OR' },
}, cb);
```

Comment (creindle) [reply services/orderMetaService/lookupDialogFormId.js]: Thank you! After this PR is done, I'll definitely want to see how you figured that out!!

File: emulation/models/orwdxm-formid.js [hunk 9]
```diff
@@ -0,0 +1,36 @@
...
+
+'use strict';
+
+const orderMetaService = require('../../services/orderMetaService');
+const modelUtils = require('./modelUtils');
+
+const RPC_EMULATOR_MODEL = [{
+    id: 'ORWDXM_FORMID',
+    name: 'ORWDXM FORMID',
+    function: 'ORWDXM^FORMID',
```
mfuroyama: This should be `FORMID^ORWDXM` <<shakes fist at empty air cursing KD's name>>

creindle:
THANK YOUUU MIKEEE!

PR #201, opened 2018-01-22
Vam 127 orwu patch

kekuso:
- fixed the property names in scripts/mvdm/MVDM-9.4-0.json


services/cprsMetaService/lookupPatch.js
```diff
@@ -0,0 +1,73 @@
...
+    // search for the specified patch in patchApplicationHistory
+    const processPackageQuery = (res, cb) => {
+        if (!res || !res.length) return cb(null, false);
+        const patchList = res;
+        const patchFound = patchList.some((patch) => {
+            const versionList = patch.version;
+            const versionObject = versionList.find(obj => parseFloat(obj.version) === parseFloat(patchVersion));
+            // first check for a matching version - if no matching version, quit early and return false
+            if (!versionObject || _.isEmpty(versionObject)) {
+                return cb(null, false);
```
mfuroyama: Correct me if I'm wrong, but I think the conditional case here should return `false` here in the `some` iteratee, instead of calling the callback. Calling the callback here might have unintended consequences.

services/cprsMetaService/lookupPatch.js
```diff
@@ -0,0 +1,73 @@
...
+        }, cb);
+    };
+
+    // search for the specified patch in patchApplicationHistory
+    const processPackageQuery = (res, cb) => {
+        if (!res || !res.length) return cb(null, false);
+        const patchList = res;
+        const patchFound = patchList.some((patch) => {
+            const versionList = patch.version;
+            const versionObject = versionList.find(obj => parseFloat(obj.version) === parseFloat(patchVersion));
```
mfuroyama: Really minor style issue, but here, consider performing the `parseFloat(patchVersion)` just once at the top level instead of doing it repeatedly in the `some` iteratee...

services/cprsMetaService/lookupPatch.js
```diff
@@ -0,0 +1,73 @@
...
+            // first check for a matching version - if no matching version, quit early and return false
+            if (!versionObject || _.isEmpty(versionObject)) {
+                return cb(null, false);
+            }
+            // check for a matching patchId and return result
+            const hasPatch = versionObject.patchApplicationHistory.some((obj) => {
+                const history = _.get(obj, 'patchApplicationHistory', '');
+                const isMatch = (parseFloat(history) === parseFloat(patchId) || (history.indexOf(`${patchId} SEQ`) !== -1));
```
mfuroyama: Same here: consider doing `parseFloat(patchId)` somewhere at the top level

PR #203, opened 2018-01-23
Vam 42 save all window size

sidjtd:
3 files and adding addChangeMultiple (and comments on how it works) in parameterService.js


emulation/models/orwch-saveall.js
```diff
@@ -0,0 +1,100 @@
...
+            const paramObj = rpcArgs[0];
+            let measureIsEmpty = false;
+            const emptyMeasureArr = [];
+
+            /* Check for non-Object params */
+            if (!_.isObject(paramObj)) callback(null, '');
+
+            const isValid = _.every(paramObj, (val, _eachKey) => {
+                if ((!_.isNumber(+_eachKey) || _.isNaN(+_eachKey)) || _eachKey < 0) return false; // all keys must be a positive number
+                if (!val.match(/^([BWC]\^).+/)) callback(null, 'all elements must begin with "B^", "W^", or "C^"');
```
mfuroyama: Ahh, so close! Calling the callback in `every` here in the iteratee function might cause some unexpected side effects.

Perhaps, instead, consider returning `false` here and setting an `errorMessage` in the scope above the `_.every` call. This gets rid of the erroneous `callback` call and preserves the error message.

emulation/models/orwch-saveall.js
```diff
@@ -0,0 +1,100 @@
...
+                const measure = val.split('^')[2];
+
+                if (measure.length !== 0) {
+                    const valArray = measure.split(',');
+                    if (val[0] === 'B' && valArray.length !== 4) return false; // B: (ORWCH BOUNDS) Bounds must have 4 values separated by commas to be valid
+                    if (!valArray.every(char => parseInt(char, 10) >= 0 && parseInt(char, 10) <= 99999)) callback(null, 'Value must be between 0 and 99999');
```
mfuroyama: Same comment here: let's replace the call to `callback` with a return of `false` and perhaps an error message variable set? (See previous comment)

mfuroyama: Also (and you don't have to change your code), but lodash does have a `_.inRange` function which checks to see if a value is between 1 or 2 other values...just something to consider for later

emulation/models/orwch-saveall.js
```diff
@@ -0,0 +1,100 @@
...
+                _.forEach(emptyMeasureArr, (val) => {
+                    paramObj[val] = '';
+                });
+            } /* End of checks */
+
+
+            const ordered = {};
+            Object.keys(paramObj).sort().forEach((key) => { // Sort object by key
+                ordered[key] = paramObj[key];
+            });
```
mfuroyama: You might want to reconsider your logic here: Although it probably works in most cases, the sort order of object keys isn't always guaranteed. Array sort orders, however, **are** guaranteed.

So, instead of populating the `ordered` object with values from a sorted key array, what you might want to consider doing is:
1) Create an array of keys in `paramObj` and sort those keys
2) Iterate over the sorted key list, then access the appropriate value in `paramObj` inside the iteratee function
3) Populate `multipleObj` accordingly

So:
```Javascript
const orderedKeys = _.keys(paramObj).sort();
orderedKeys.forEach((key) => {
    val = paramObj[key];
    ...
});
```

emulation/models/orwch-saveall.js
```diff
@@ -0,0 +1,100 @@
...
+                };
+
+                if (label) multipleObj.push(model);
+            });
+
+            ParameterService.withContext(session).addChangeMultiple(multipleObj, callback)
+                .then(res => callback(null))
+                .catch(err => callback(err));
+        },
+        formatResult: res => res,
```
mfuroyama: Let's remove `formatResult`. It's not doing anything, we can leave it

emulation/models/orwch-saveall.js
```diff
@@ -0,0 +1,103 @@
...
+                    if (val[0] === 'B' && valArray.length !== 4) return false; // B: (ORWCH BOUNDS) Bounds must have 4 values separated by commas to be valid
+                    if (!valArray.every(char => parseInt(char, 10) >= 0 && parseInt(char, 10) <= 99999)) {
+                        errorMessage = 'Value must be between 0 and 99999';
+                        return false;
+                    }
+                } else {
+                    emptyMeasureArr.push(_eachKey);
+                    measureIsEmpty = true; // check for empty measurement values
+                }
+                if (errorMessage) callback(null, errorMessage);
```
mfuroyama: A good rule of thumb: if you're in an iteratee function, you probably shouldn't be calling the `callback` function.

This is a good start, but you might want to set the initial `errorMessage` value to empty string (`''`), then after the call to `_.every`, in your check on `isValid`, do a:

```Javascript
if (!isValid) return callback(null, errorMessage);
```

Also, don't forget that you should probably `return callback` below...

sidjtd: OMG YOU ARE RIGHT! I thought I placed this right outside the _every like you suggested in the last comments and its still within it! Yes I'll move it outside....

PR #206, opened 2018-01-24
Vam-26: GMV Parameter Not Stashed

creindle:
[X] Async Query Functions!
[X] Passes comparison, model, doc services, and services tests


services/cprsMetaService/gmvParameter.js
```diff
@@ -43,86 +44,109 @@ function gmvParameter(routine, entArg, gmvPrmArg, instArg, valArg, context, cb)
...
 
     // 'ENTVAL' load domain info via entity value
-    } else if (routine === 'ENTVAL') {
-        if (entArg === 'USR') cb(null, context.name);
-        else if (entArg === 'SYS') {
-            this.find('KernelSystemParameters', {
+    function ENTVALRoutine(cb) {
+        if (entArg === 'USR') return cb(serviceUtils.bail(context.name));
+        else if (entArg === 'DIV') {
+            return cb(serviceUtils.bail({ srv: context.division.label, key: context.division.label }));
```
creindle: I am not 100% sure if this should be `return cb(serviceUtils.bail({ srv: context.division.label, key: context.division.label }));` or `cb(null, { srv: context.division.label, key: context.division.label }` because the original implementation just had it return the object

emulation/models/gmv-parameter.js
```diff
@@ -0,0 +1,97 @@
...
+            'GMV WEBLINK',
+            'GMV ALLOW USER TEMPLATES',
+            'GMV DEFAULT VALUES',
+            'GMV TEMPLATE',
+            'GMV TEMPLATE DEFAULT',
+            'GMV USER DEFAULTS',
+        ],
+        files: ['8989.3', '8989.51', '8989.518'],
+        domain: 'CPRS',
+        labels: ['META', 'VITAL', 'CHANGE'],
```
creindle: This did not affect the DB, so did not need to add the comment 'RESULTS-UNCHECKED'

services/cprsMetaService/gmvParameter.js
mfuroyama: It should be:
```Javascript
return cb(serviceUtils.bail({ srv: context.division.label, key: context.division.label }));
```

The reason this works is because you're causing the function chain to break out by passing in an error, which the `multipleQueries` function deals with at the back end.

services/cprsMetaService/gmvParameter.js
```diff
@@ -43,86 +44,109 @@ function gmvParameter(routine, entArg, gmvPrmArg, instArg, valArg, context, cb)
     let finalOutput = false;
```
mfuroyama: This is fine as it is, no need to make changes, but here's a "best practice" that I like to follow regarding the variable declaration here:

Most of the parameters above are **constants**, so what I like to do with these normally is **capitalize** the variable names, and then move them into the top-most scope.

The capitalization really helps with readability, because if you see a capitalized variable being used, immediately, you can assume that this value is constant.

services/cprsMetaService/gmvParameter.js
```diff
@@ -43,86 +44,109 @@ function gmvParameter(routine, entArg, gmvPrmArg, instArg, valArg, context, cb)
...
+            }, cb);
         }
+    }
+    function ENTVALProcess(res, cb) {
+        if (entArg === 'SYS') {
+            if (_.get(res[0], 'domainName.label', '')) finalOutput = res[0].domainName;
+            return cb(null, finalOutput);
+        } else if (entArg.match(ENTITY_REGEX)) {
+            cb(null, res[0] || {});
+        }
```
mfuroyama: I think we've talked about this before, but this is potentially dangerous. The way the function is set up here, if `entArg` is not equal to 'SYS' and `entArg.match(ENTITY_REGEX)` is false, the callback `cb` will never get called which is **very** not good.

On a "best practices" related note, it's usually good form to **not** end a function in a conditional case.

services/cprsMetaService/gmvParameter.js
```diff
@@ -43,86 +44,110 @@ function gmvParameter(routine, entArg, gmvPrmArg, instArg, valArg, context, cb)
...
+            }, cb);
         }
+        return cb(serviceUtils.bail(new Error('Incorrect entArg')));
+    }
+    function ENTVALProcess(res, cb) {
+        if (entArg === 'SYS') {
+            if (_.get(res[0], 'domainName.label', '')) finalOutput = res[0].domainName;
+            return cb(null, finalOutput);
+        }
+        const finalResponse = entArg.match(ENTITY_REGEX) && !_.isEmpty(res) ? res[0] : {};
```
mfuroyama: Perfect! 👍

PR #210, opened 2018-01-26
Vam 365 orqpt default patient list

pcchanghawaii:
I'm guess the default patient list comes from the following field in NewPerson file.

101.02  patient selection list

All (of our test) users return empty array in the RPC calls' results.


emulation/models/orqpt-default-patient-list.js
```diff
@@ -0,0 +1,30 @@
...
+    function: 'DEFLIST^ORQPTQ11',
+    domain: 'USER/PATIENT',
+    description: 'Return user’s default patient list source.',
+    returnType: 'ARRAY',
+    rpcType: 'READ',
+    complexity: 'LOW',
+    files: [],
+    parameters: [],
+    labels: ['META', 'USER-NOT-REQUIRED', 'HARDCODED'],
+    services: [],
```
creindle: The value for the key `services` should be the name of the service function being used in this model, which is `userMetaService.lookupDefaultPatientList`.

emulation/models/orqpt-default-patient-list.js
```diff
@@ -0,0 +1,30 @@
...
+    inputParameters: [],
+    invokeRPC: (args, session, callback) => {
+        userMetaService.lookupDefaultPatientList(session, callback);
+    },
+
+    formatResult: (res) => {
```
creindle: Since this formatResult is not changing the output of `res`, we can remove this method and ensure the result of the service function is `res[0]`.

emulation/models/tests/orqpt-default-patient-list-spec.js
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env node
```
creindle: We could then remove this file since it is testing the removed method `formatResults` that was originally in the model file `orqpt-default-patient-list.js`.

services/userMetaService/lookupDefaultPatientList.js
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * @memberof UserMetaService
+ * @instance
+ */
+
+function lookupDefaultPatientList(context, callback) {
+
+    this.find('NewPerson', {
+        query: { id: context.userId, inactiveFlag: { $exists: false }, localStationNumber: context.stationId },
+        //projection: serviceUtils.createProjection(['restrictPatientSelection', 'patientSelectionList']),
```
creindle: Were you planning on removing this projection? Or was this up for discussion? Just wanted to double check!

services/userMetaService/lookupDefaultPatientList.js
```diff
@@ -0,0 +1,46 @@
...
+        projection: serviceUtils.createProjection(['patientSelectionList']),
+    })
+    .then((elementArr) => {
+            return callback(null, elementArr);
```
creindle: We can actually remove the `return` statement because we are not breaking out of a block statement. `callback(null, elementArr)` would be enough!

services/userMetaService/lookupDefaultPatientList.js
```diff
@@ -0,0 +1,46 @@
...
+    })
+    .catch((err) => {
+            fail(err);
```
creindle: We can remove the `fail(err)` here because we would be testing for them in the spec file and would give us some trouble if the service stopped completely because of the fail.

services/userMetaService/lookupDefaultPatientList.js
```diff
@@ -0,0 +1,46 @@
...
+            return callback(err);
```
creindle: And we can remove the `return` here too! And simply leave `callback(err)`!

services/userMetaService/lookupDefaultPatientList.js
```diff
@@ -0,0 +1,46 @@
...
+    });
+/*
```
creindle: Did you want to leave this section? I could ask @mfuroyama what the best direction should be for your comments in the PR.

pcchanghawaii:
if I'm right on the guess of patient list, then we need to import this file  OE/RR LIST (100_21) as well in order to get the patient list, if any.  This RPC may need further modification. It's hard coded as empty/undefined list for now

mfuroyama:
Hey Pei! Per our discussion, I'm going to move this in with the promise that we'll revisit this in the future.

PR #211, opened 2018-01-26
orqpt default list sort added

pcchanghawaii:
hard code for now.  still trying to figure out which file/field to get the RPC call result.

45.7 may be related
OPTION (19) may be related.


emulation/models/orqpt-default-list-sort.js
```diff
@@ -0,0 +1,30 @@
...
+    function: 'DEFSORT^ORQPTQ11',
+    domain: 'PATIENT',
+    description: 'Returns the current user’s default patient selection list SORT ORDER setting.',
+    returnType: 'SINGLE VALUE',
+    rpcType: 'READ',
+    complexity: 'LOW',
+    files: [],
+    parameters: [],
+    labels: ['META', 'USER-NOT-REQUIRED', 'HARDCODED'],
+    services: [],
```
mfuroyama: Hey Pei!

So sorry...but I'm finally getting to the pull requests, now that things have calmed down a little.

Anyway, you've probably seen from comments from the team, that Conor wants us to put the **service** calls in the `services` array of the RPC models. So here, let's add: 
```
userMetaService.lookupDefaultListSort
```

emulation/models/orqpt-default-list-sort.js
```diff
@@ -0,0 +1,30 @@
...
+    inputParameters: [],
+    invokeRPC: (args, session, callback) => {
+        userMetaService.lookupDefaultListSort(session, callback);
+    },
+
+    formatResult: (res) => {
+        return res;
+    },
```
mfuroyama: Also, the inclusion of `formatResult` is optional. If your `formatResult` function doesn't do anything (i.e. just returns `res`, you can leave it out of the model. The emulator does the checks internally, so it's all good!

services/userMetaService/lookupDefaultListSort.js
```diff
@@ -0,0 +1,26 @@
...
+ * @return {String} A or D
+ *
+ * @memberof UserMetaService
+ * @instance
+ */
+
+function lookupDefaultListSort(context, callback) {
+    try {
+        const res = 'A'; // default to ascending
+        // const userId = parseInt(context.userId.split(':')[4], 10); // urn:vista:6100:200:63
```
mfuroyama: I know this is commented out, but you might want to consider using the `serviceUtils.getIdFromUrn()` function here, which extracts the "IEN" portion out of the URN.

PR #212, opened 2018-01-26
Vam 273 do management tasks

sidjtd:
GMV Manager, 6 functions and a default method.


services/vitalsMetaService/doManagementTasks.js
```diff
@@ -0,0 +1,304 @@
...
+            ParameterService.withContext(_context).get('GMV TEMPLATE DEFAULT', { entity: 'SYS' })
+                .then((result) => {
+                    const res = _.isEmpty(result) ? '-1^No Default Template' : result;
+                    return _cb(null, {
+                        option: 'GETDEF',
+                        result: res,
+                    });
+                }).catch(err => _cb(err));
+        }
+        ParameterService.withContext(_context).get('GMV TEMPLATE DEFAULT', { entity: 'ALL' })
```
mfuroyama: Whoever wrote this: Did you intend to call the second Parameter Service get call wrapped in an `else` block? If not, then this code will potentially create problems, since there's a possibility that `_cb` gets called twice.

I would fix it myself, but the intent here is not clear

services/vitalsMetaService/doManagementTasks.js
```diff
@@ -0,0 +1,304 @@
...
+     *                               3. EntityLabel (i.e. 'DEMO.NODEVISTA.ORG')
+     *                               4. Instance (i.e. 'DAILY VITALS')
+     *                               5. Value (i.e. 'TPR, BP|1:0;5:0;3:0;2:0;21:0;8:0;22:0;9:0')
+     *                            OR [0] if the user does not have the right permissions
+     */
+    GETDATA: (_data, _service, _context, _cb) => {
+        const option = 'GETDATA';
+        /* _data arrives in following order: {File Type}^{IEN}^{Field Number} */
+        const component = _data.split('^');
+        if (component.length < 3) _cb(null, []);
```
mfuroyama: `_cb(null, [])` should be `return _cb(null, [])`

emulation/models/gmv-manager.js
```diff
@@ -0,0 +1,98 @@
...
+            }
+            if (_.isEqual(getType, 'GETTEMP')) {
+                if (res.error) return res.error;
+                const finalRes = res.result.map((val) => {
+                    if (val.size) return val.size;
+                    const entityId = convertId(val.entityType, val.entityId);
+                    return `${ENTITY_TYPE_MAP[val.entityType]}^${entityId}^${val.entityLabel}^${val.instance}^${val.value}`;
+                });
+                return finalRes;
+            }
```
mfuroyama: A map object with function fields might be a better choice here

services/vitalsMetaService/doManagementTasks.js
```diff
@@ -0,0 +1,304 @@
...
+                    instance: valObj.instance,
+                    value: valObj.value,
+                });
+            });
+            cback(null, results);
+        };
+
+        const queryGmvAllowUserTemplates = (res, cback) => {
+            const isUserEntity = res.some((obj => obj.entityType === 'USR'));
+            if (isUserEntity) ParameterService.withContext(context).get('GMV ALLOW USER TEMPLATES', { entity: 'SYS' }, cb);
```
mfuroyama: Correct me if I'm mistaken, but I think the `cb` here should be `cback`?

services/vitalsMetaService/doManagementTasks.js
```diff
@@ -0,0 +1,304 @@
...
+            service.find('NewPerson', {
+                query: { id: context.userId },
+                projection: serviceUtils.createProjection(['keys.key.label']),
+            }, cback);
+        };
+
+        const processKeysQuery = (res, cback) => {
+            const key = 'GMV MANAGER';
+            if (!res || !res.length) return cback(null, appendOption(option, [0]));
+            const keys = res[0] ? _.get(res[0], 'keys', '') : '';
```
mfuroyama: This can be reduced to:
```Javascript
const keys = _.get(res[0], 'keys', '');
```
The `_.get` guards against a potentially invalid `res[0]` value

services/vitalsMetaService/doManagementTasks.js
sidjtd: I think that was christie, I’m gonna decipher it now

sidjtd: Yup, in the original code there is an implied "else", as the 'if statement' ends with a return.  I'll wrap it in an else to avoid the risk of the _cb call being called twice

PR #214, opened 2018-01-27
Vam 367 orwpt id info

pcchanghawaii:
wardLocation is used for Ward field in this implementation. It could be lodgerWardLocation as well.


emulation/models/orqpt-default-list-sort.js
```diff
@@ -0,0 +1,30 @@
...
+    function: 'DEFSORT^ORQPTQ11',
+    domain: 'PATIENT',
+    description: 'Returns the current user’s default patient selection list SORT ORDER setting.',
+    returnType: 'SINGLE VALUE',
+    rpcType: 'READ',
+    complexity: 'LOW',
+    files: [],
+    parameters: [],
+    labels: ['META', 'USER-NOT-REQUIRED', 'HARDCODED'],
+    services: [],
```
creindle: We need to add the service function being used, so we would add `userMetaService.lookupDefaultListSort` as the value for the key `services`.

emulation/models/orqpt-default-list-sort.js
```diff
@@ -0,0 +1,30 @@
...
+    inputParameters: [],
+    invokeRPC: (args, session, callback) => {
+        userMetaService.lookupDefaultListSort(session, callback);
+    },
+
+    formatResult: (res) => {
```
creindle: Since we are only returning `res` without changing its format, we can simply remove this method entirely!

emulation/models/orwpt-id-info.js
```diff
@@ -0,0 +1,49 @@
...
+        files: ['2'],
+        parameters: [{
+            id: 'DFN',
+            parameterType: 'LITERAL',
+            required: true,
+            description: 'Patient IEN',
+            example: '11',
+        }],
+        labels: ['META', 'hardcoded'],
+        services: ['patientMetaService'],
```
creindle: Before we had only the meta service as the value for services, but now we can add the service function being used! So we would change this to `patientMetaService.lookupPatientInfo`.

emulation/models/orwpt-id-info.js
```diff
@@ -0,0 +1,49 @@
...
+        },
+        formatResult: res => {
+            let ret = '';
+            if (res) {
+                ret = Object.keys(res).map(key => res[key]).join('^');
+                
+            }       
+            return ret;
+        }
+        //res.map(element => `${serviceUtils.getIdFromUrn(element.subjectAlternativeName)}^${element.accessCode}`),
```
creindle: Did you want to leave this in for discussion? This looks important so wanted to check!

emulation/models/tests/orqpt-default-list-sort-spec.js
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env node
```
creindle: We can remove this file since this is testing `formatResult` in the model file and we are removing that method!

emulation/models/tests/orwpt-id-info-spec.js
```diff
@@ -0,0 +1,31 @@
...
+
+    beforeAll(() => {
+        const res = unitTestUtils.beforeAllScript(rpcEmulatorModel);
+        rpc = res;
+        context = res.context;
+    });
+
+    it('formatResult: Should return specfic information about a patient', () => {
+        let res;
+        for (let i = 0; i < testInput.length; i++) {
```
creindle: Hey Pei! So for this, we use an ESLint spec that discourages the use of for loops. It's up to you, but you could use a `forEach` loop instead to prevent that error! I apologize for a trivial matter...

services/patientMetaService/lookupPatientInfo.js
```diff
@@ -0,0 +1,33 @@
...
+ * @return {string} results of patient information  000000113^2810302^M^Y^^^^CARTER,DAVID
+ *                  PID^DOB^SEX^VET^SC%^WARD^RM-BED^NAME
+ *                  service connected %    room-bed
+ * @memberof patientMetaService
+ * @instance
+ */
+function lookupPatientInfo(patientId, context, callback) {
+    this.find('Patient', { // needs to be updated..
+        //query: { id: `200-${patientId}`/*, localStationNumber: context.stationId*/ },
+        query: { id: patientId },
```
creindle: Did you want to remove line 17 and insert a query for `localStationNumber` here? Then, we can remove the commented out code.

services/tests/lookupDefaultListSort-spec.js
```diff
@@ -0,0 +1,28 @@
...
+
+describe('lookupDefaultPatientList asynchronous specs', () => {
+    let context;
+    beforeAll(() => {
+        context = serviceTestUtils.beforeAllScript(userMetaService).context;
+    });
+
+
+    const testDefault = (res) => {
+        if (res === 'A') {
```
creindle: I think I'm a little confused at this part, if `res` is an Object, is 'A' a specific value of one of the properties of the object being passed?

services/userMetaService/lookupDefaultListSort.js
```diff
@@ -0,0 +1,26 @@
...
+ * @instance
+ */
+
+function lookupDefaultListSort(context, callback) {
+    try {
+        const res = 'A'; // default to ascending
+        // const userId = parseInt(context.userId.split(':')[4], 10); // urn:vista:6100:200:63
+        return callback(null, res);
+    } catch (err) {
+        fail(err);
```
creindle: You can remove this line because we don't want the service function to fail here! Passing `err` as a callback is enough!

services/userMetaService/lookupDefaultListSort.js
```diff
@@ -0,0 +1,26 @@
...
+        return callback(err);
```
creindle: We can remove the `return` at this part, to simply have `callback(err)`.

services/tests/lookupPatientInfo-spec.js
```diff
@@ -0,0 +1,80 @@
...
+              if (res[0].roombed)
+                expect(res[0].roombed).toEqual(jasmine.any(String));
+            }
+            done();
+        });
+    });
+
+    it('Should not return info for a non existent patient', (done) => {
+        service.lookupPatientInfo(499, context, (err, res) => {
+            if (err) {
```
creindle: We can reduce these lines of code to `if (err) fail(err)`!

services/tests/lookupPatientInfo-spec.js
```diff
@@ -0,0 +1,80 @@
...
+                expect(res[0].lodgerWardLocation).toEqual(jasmine.any(String));
+              if (res[0].roombed)
+                expect(res[0].roombed).toEqual(jasmine.any(String));
+            }
+            done();
+        });
+    });
+
+    it('Should return info for a non standard patient', (done) => {
+        service.lookupPatientInfo('63', context, (err, res) => {
```
creindle: A comment @mfuroyama has made on one of my PRs before is that when we are testing or creating service functions, we should be passing in entire URNs to the function, as service functions have no concept of IENs. This guideline was written out (here)[https://github.com/vistadataproject/VICSServer/wiki/Guidelines-for-VICS-Services] by Conor

services/tests/lookupPatientInfo-spec.js
```diff
@@ -0,0 +1,80 @@
...
+            if (err) {
```
creindle: We can reduce these lines of code to if (err) fail(err)!

services/userMetaService/lookupDefaultListSort.js
kekuso: Yes, I would remove line 19.

pcchanghawaii:
https://github.com/vistadataproject/VICSServer/tree/vam-529-orwpt-id-info
please refer to this new pull request due to the updated Patient collection.

PR #215, opened 2018-01-31
Create cprs router client

kekuso:
Directions: 
1. Make sure the VICSServer emulation is running (`node index.js` in `VICSServer/emulation/server`), Rpc router is running (`node router.js` in `rpcRouter/`).
2. run `node rpcRouterClient.js` in `VICSServer/examples` with optional command line args `-h [host] -p [port)]`

Just having some issues with the connection to nodevista (-h 10.2.2.100 -p 9430), getting a console output `)Application context has not been created!` (Maybe something to do with userIEN not matching?)


examples/rpcClient.js
```diff
@@ -0,0 +1,115 @@
+'use strict';
+
+const net = require('net');
+
+const Promise = require('promise');
```
mfuroyama: Let's use 'bluebird' here, since it's the Promise library that we're using in the main VICS Server project. Bluebird is a faster, more efficient drop-in replacement for the native NodeJS Promise library:
```Javascript
const Promise = require('bluebird');
```

package.json
```diff
@@ -47,9 +47,11 @@
     "fs-extra": "^5.0.0",
     "lodash": "^4.17.4",
     "mathjs": "^3.16.3",
+    "minimist": "^1.2.0",
     "moment": "^2.18.1",
     "mongoose": "^4.11.7",
     "progress": "^2.0.0",
+    "promise": "^8.0.1",
```
mfuroyama: You can remove this

examples/rpcRouterClient.js
```diff
@@ -0,0 +1,232 @@
...
+
+const PROBLEM_TEST_CASE = {
+    search: 'ACHALASIA',
+    view: 'PLS',
+    date: '3050701',
+    synonyms: true,
+    completion: true,
+};
+
+const genericCallback = (err, res, fulfill, reject) => {
```
mfuroyama: Actually, from the looks of Steve's code, you don't need this generic callback. In fact, from the way you're calling the code (using Promise chains), you **definitely** want to remove the callback parameters in the calls. If you use both the Promise **and** the callbacks, you may end up the Promise resolver functions twice, which is "not good"

examples/rpcRouterClient.js
```diff
@@ -0,0 +1,232 @@
...
+        if (err) {
+            reject(err);
+        } else if (res === rpcFormatter.encapsulate('accept')) {
+            fulfill(res);
+        } else {
+            reject(new Error('Greeting error (TCP Connect)'));
+        }
+    };
+    // arguments of msg and the client's IP address
+    return Client.sendRpc(rpcFormatter.buildRpcGreetingString(Client.getClient().localAddress, msg), callback);
```
mfuroyama: So, for example, here, you don't need to define the`callback` function, and you can remove the `callback` parameter to the `Client.sendRpc` call here. Just rely on the Promise mechanism and don't bother with the callbacks

mfuroyama:
@kekuso: I pushed a super simple test client that I hacked together really quickly. This client seems to work against both the RPC Router and the native RPC Broker port of nodeVISTA, using the sequence of RPCs used in the performance test setup. I checked my stuff into `examples/simpleTestClient.js`

Check out the sequences and the raw data and use that to double-check the inputs and outputs of your client to see why stuff might not be working.

PR #219, opened 2018-02-02
VAM-539: Vitals READ functions

creindle:
[X] Implemented "list" and "lookup" functions
[X] Implemented converter from Fileman time to ISO


services/VitalsService/listVitals.js
```diff
@@ -0,0 +1,60 @@
+'use strict';
+
+const _ = require('lodash');
+const moment = require('moment');
+const convertFMtoISO = require('../dateTimeService/convertFMtoISO').convertFMtoISO;
```
mfuroyama: Here, let's just get the `dateTimeService` then call the `convertFMtoISO` function from it

services/VitalsService/listVitals.js
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * @summary Returns list of vitals objects associated with context object "patientId"
+ * @param  {Number}  startDate  Start date/time for vital retrieval in Fileman format.\rIf none is passed then the start date is 1 (i.e. before any dates)
+ * @param  {Number}  endDate  Stop date/time for vital retrieval in Fileman format.\rIf none is passed then the Start date is also the stop date and if there\ris not start date then 9999999 is used as the stop date
```
mfuroyama: Can we change the function to accept Date objects instead of Fileman-formatted date/time? We want the service functions to be as free from the smell of Fileman as possible (no IENs, no Fileman dates)

services/VitalsService/listVitals.js
```diff
@@ -0,0 +1,60 @@
...
+    let isoStartDate = convertFMtoISO(startDate);
+    let isoEndDate = convertFMtoISO(endDate);
+    if (!isoEndDate && !isoStartDate) isoEndDate = moment().format('YYYY-MM-DDTHH:mm:ssZ'); // current date time
+    if (!startDate) isoStartDate = moment('1900-01-01T00:00:00').format('YYYY-MM-DDTHH:mm:ssZ'); // by default push date out to before records were collected
+    if (!isoEndDate) isoEndDate = isoStartDate;
+    const vitals = [];
+    this.find('Vital', {
+        query: { 'patient.id': context.patientId, localStationNumber: context.stationId },
+    }).then((res) => {
+        _.forEach(res, (vital) => {
```
mfuroyama: This works, but consider using `filter` here instead of `forEach`; using `filter` signals your intent, and makes for slightly cleaner code.

Also, when possible, use the ES6 built-in `Array` iterator functions. So, unless you're going to do something that takes advantage of lodash's special functionality, prefer `res.filter` over `_.filter(res, ...)`

services/VitalsService/listVitals.js
```diff
@@ -0,0 +1,60 @@
...
+            let vitalTakenDateTime = _.get(vital, 'vitalsTakenDateTime', '');
+            vitalTakenDateTime = moment(vitalTakenDateTime).format('YYYY-MM-DDTHH:mm:ssZ');
+            // check if vital taken time is in between the input start and end dates
+            if (vitalTakenDateTime >= isoStartDate && vitalTakenDateTime <= isoEndDate) {
```
mfuroyama: So here, instead of using Moment.js to convert times to strings and doing the comparisons here, I'd suggest just converting all the dates to Moment.js objects, and taking advantage of the Moment functionality to do the comparison. Moment.js has an [isBetween](http://momentjs.com/docs/#/query/is-between/) function that you can take advantage of:
```Javascript
const momentStartDate = moment(startDate);
const momentEndDate = moment(endDate);
...
// Since you'll be doing a "filter":
const momentTaken = moment(vitalTakenDateTime);
return momentTaken.isBetween(momentStartDate, momentEndDate);
```

Your code will not only end up more reliable, but easier to read as well

services/VitalsService/listVitals.js
```diff
@@ -0,0 +1,60 @@
...
+                vitals.push(vital);
+            }
+        });
+        // grab most recent
+        const mostRecentByType = {};
+        _.forEach(vitals, (vital) => {
```
mfuroyama: Instead of the implementation here in the `_.forEach`, consider this approach:

* First, before you do anything, sort the `vitals` list by `vitalsTakenDateTime`, latest to earliest
* Then, take take advantage of lodash's [`groupBy`](https://lodash.com/docs/4.17.4#groupBy) function to group the arrays by `vitalType.label` in an object:
```Javascript
const vitalsByType = _.groupBy(dateSortedVitals, 'vitalType.label');
```
* Then, `map` the keys from the `vitalsByType` object and take the first item off of the list corresponding to the type:
```Javascript
const mostRecentList = _.keys(vitalsByType).map(type => _.first(vitalsByType[type]));
```

This might be confusing, so let's talk about it if you want

services/VitalsService/lookupVitals.js
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * @memberof VitalsService
+ * @instance
+ */
+
+function lookupVitals(vitalId, context, callback) {
+    this.find('Vital', {
+        query: { id: vitalId, 'patient.id': context.patientId, localStationNumber: context.stationId },
+    }).then((vitalObj) => {
+        const response = vitalObj[0] || [];
```
mfuroyama: Might be better here to default to an empty object instead of an empty array to keep the return types consistent.

services/dateTimeService/convertFMtoISO.js
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * @return {String} convertedISODate An ISO 8601 date to convert
+ * @example
+ *  '2017-05-05T06:27:21-10:00'
+ *
+ * @memberOf  DateTimeService
+ * @instance
+ */
+function convertFMtoISO(fmDate) {
+    if (!fmDate || _.isNaN(_.toNumber(fmDate))) {
```
mfuroyama: Hmm, since you're requiring `fmDate` to be a string, I'd say this guard might be better served if you replaced the `!fmDate` check with an `!_.isString(fmDate)` check

services/dateTimeService/convertFMtoISO.js
```diff
@@ -0,0 +1,40 @@
...
+        return null;
+    }
+    const dateStr = _.split(fmDate, '.')[0] || '';
```
mfuroyama: I'd suggest just doing the `split` here once, using the built-in String `split` function and using the output of that in your function. No need to do the `split` twice:
```Javascript
const parts = fmDate.split('.');
const dateStr = parts[0];
```

services/dateTimeService/convertFMtoISO.js
```diff
@@ -0,0 +1,40 @@
...
+    const monthDayStr = dateStr.substring(dateStr.length - 4, dateStr.length);
+    let yearStr = dateStr.substring(0, dateStr.length - 4) || '';
+    let timeStr = _.split(fmDate, '.')[1] || '';
+    if (timeStr) {
```
mfuroyama: Actually, when you think about it, you don't really need to do this, since what you're doing is just breaking up the string into two-character substrings, then pasting them together in the same order

services/dateTimeService/convertFMtoISO.js
```diff
@@ -0,0 +1,40 @@
...
+        const hrStr = timeStr.substring(0, 2) || '';
+        const minStr = timeStr.substring(2, 4) || '';
+        const secStr = timeStr.substring(4, 6) || '';
+        timeStr = ` ${hrStr}${minStr}${secStr}`;
+    }
+    if (yearStr) yearStr = (_.toNumber(yearStr) + 2000) - 300;
+    const convertedISODate = moment(_.toString(yearStr) + monthDayStr + timeStr).format('YYYY-MM-DDTHH:mm:ssZ');
```
mfuroyama: Here, let's use Javascripts built-in `toString()` function instead of the lodash version.

mfuroyama: Also, since we know the format of the date/time string we're passing into the `moment` constructor, let's pass it in as a string:
```Javascript
const dateTimeStr = `${yearStr}${monthDayStr} ${timeStr}`;
const convertedISODate = moment(dateTimeStr, 'YYYYMMDD HHmmss').format(...);
```

services/tests/listVitals-spec.js
```diff
@@ -0,0 +1,20 @@
...
+const serviceTestUtils = require('./serviceTestUtils');
+
+describe('listVitals asynchronous specs', () => {
+    let context;
+    beforeAll(() => {
+        context = serviceTestUtils.beforeAllScript().context;
+    });
+
+    it('Invoking function with start and end time should return array of vital objects', (done) => {
+        VitalsService.listVitals(3170505, 3180517.062721, context, (err, res) => {
```
mfuroyama: Don't forget to change these out from Fileman formatted date/time values

mfuroyama: Also, you might want to alter the `context` object so that your stationId is `999`, which will ensure that you get the test data

services/VitalsService/listVitals.js
creindle: Perfect!

services/dateTimeService/convertFMtoISO.js
creindle: That's... a good point...!

services/VitalsService/listVitals.js
```diff
@@ -0,0 +1,50 @@
+'use strict';
+
+const _ = require('lodash');
+const moment = require('moment');
+/**
+ * @summary Returns list of vitals objects associated with context object "patientId"
+ * @param  {Object}  startDate  Start date/time for vital retrieval in ISO format
+ * @param  {Object}  endDate  Stop date/time for vital retrieval in ISO format.
```
mfuroyama: Let's change the data types of `startDate` and `endDate` to `String` here

services/VitalsService/listVitals.js
```diff
@@ -0,0 +1,50 @@
...
+            // check if vital taken time is in between the input start and end dates
+            return momentTaken.isBetween(momentStartDate, momentEndDate);
+        });
+        // _.sortBy() sorts the array earliest to latest, but we want it by latest first
+        const dateSortedVitals = _.sortBy(inputtedVitals, 'vitalsTakenDateTime').reverse();
+        // grab most recent
+        const mostRecentByType = {};
+        const vitalsByType = _.groupBy(dateSortedVitals, 'vitalType.label');
+        const mostRecentList = _.keys(vitalsByType).map(type => _.first(vitalsByType[type]));
+        _.keys(mostRecentByType).forEach((vitalType) => {
```
mfuroyama: Let's talk about this code here...I'm a little confused as to what the intent is

creindle: Actually, I believe I can delete this line because the last line was creating the mostRecentList already.

PR #223, opened 2018-02-09
Vam 505 orqqal detail

kekuso:
Let me know if there's a better (mutator) version of concat that I can use! Also, tried lodash chaining with _.get and _.startCase but couldn't get it working for some reason...you'll see what I mean.


emulation/models/orqqal-detail.js
```diff
@@ -0,0 +1,95 @@
...
+        name: 'ORQQAL DETAIL',
+        function: 'DETAIL^ORQQAL',
+        domain: 'ALLERGY',
+        description: 'Returns detailed allergy info for specified allergic reaction.',
+        rpcType: 'READ',
+        complexity: 'LOW',
+        returnType: 'ARRAY',
+        files: ['120.8'],
+        parameters: [],
+        labels: [],
```
mfuroyama: Let's add `'PATIENT'` here to show that this is a clinical (patient) RPC

emulation/models/orqqal-detail.js
```diff
@@ -0,0 +1,95 @@
...
+            }
+
+            // add originator names and dates and verification/observation dates
+            formattedRes = formattedRes.concat([
+                `         Originator: ${res.originator.label}`,
+                `         Originated: ${res.originated}`,
+                ` Obs dates/severity: ${res.observedDatesSeverity}`,
+                '',
+                `           Verified: ${res.verified ? '<auto-verified>' : ''}`,
+                `Observed/Historical: ${_.startCase(res.observedHistorical)}`,
```
mfuroyama: This is fine

emulation/models/orqqal-detail.js
```diff
@@ -0,0 +1,95 @@
...
+                '',
+            ];
+
+            // add signs and symptoms to the formatted result, if any are present
+            if (res.signsAndSymptoms && res.signsAndSymptoms.length) {
+                // eslint-disable-next-line
+                const signsAndSymptoms = res.signsAndSymptoms.map((sign, i) => {
+                    return (i === 0) ? `       Signs/symptoms: ${sign.label}` : `                     ${sign.label}`;
+                });
+                formattedRes = formattedRes.concat(signsAndSymptoms);
```
mfuroyama: Instead of using `concat`, consider using `Array.prototype.push.apply` to modify the `formattedRes` array in-place:
```Javascript
Array.prototype.push.apply(formattedRes, signsAndSymptoms);
```

PR #224, opened 2018-02-09
vam-502-or-get-combat-vet

pcchanghawaii:
update or-get-combat-vet based on reviews, sync specs with master's Patient collection.
This pull request replaces vam-368-or-get-combat-vet as the 368 was created when Patient collection (and many other files) were not loaded .


emulation/models/or-get-combat-vet.js
```diff
@@ -0,0 +1,48 @@
...
+    rpcType: 'READ',
+    complexity: 'LOW',
+    returnType: 'ARRAY',
+    files: ['2'],
+    parameters: [{
+        id: 'DFN', // in URN
+        parameterType: 'LITERAL',
+        required: true,
+        description: 'Patient URN',
+        example: 'urn:vista:999:2:25',
```
mfuroyama: Hey Pei! Can we change this back to IENs? Unfortunately, even though the service functions should only deal in URNs, the RPC Emulation Models will continue to receive IENs from CPRS et al.

In general, the pattern is:
* RPC Emulation Model `invokeRPC`: Receive IENs in args -> Convert to URNs -> Pass to Service function
* Service Functions: Process URNs -> Return URNs
* RPC Emulation Model `formatResult`: Receive URNs in results -> Convert any URNs to IENs as necessary

I know, it seems like a lot of unnecessary work, but it helps us conceptually isolate the VICS services from the RPC emulation layer. Thanks!

emulation/models/or-get-combat-vet.js
```diff
@@ -0,0 +1,48 @@
...
+    }],
+    labels: ['META'],
+    services: ['patientMetaService.lookupPatientCombatVet'],
+    inputParameters: [],
+    invokeRPC: (args, session, callback) => {
+        patientMetaService.lookupPatientCombatVet(args, session, callback);
```
mfuroyama: So yeah, here, we should convert the first argument in the array to a URN from IEN.

Also, it looks like the `lookupPatientCombatVet` function accepts a *String* as the first function argument, but here you're passing in an array. We can probably fix this too alongside the URN-to-IEN conversion

emulation/models/or-get-combat-vet.js
```diff
@@ -0,0 +1,48 @@
...
+    },
+
+    formatResult: (res) => {
+        let ret = 'NOTCV';
+        if (res[0].militaryService) {
+            if (res[0].militaryService.combatService) {
+                if (res[0].militaryService.combatService.serviceIndicated) {
+                    ret = 'CV';
+                } else if (res[0].militaryService.combatService.combatVeteranEndDate) {
```
mfuroyama: Instead of the nested `if` statements, if you're willing to use lodash, consider using the `_.has` lodash function:
```Javascript
if (_.has(res[0], 'militaryService.combatService.serviceIndicated')) {
    ret = 'CV';
} else if (_.has(res[0], 'militaryService.combatService.combatVeteranEndDate')) {
    ret = 'CV';
} 
```

emulation/models/or-get-combat-vet.js
```diff
@@ -0,0 +1,48 @@
...
+    id: 'OR_GET_COMBAT_VET',
+    name: 'OR GET COMBAT VET',
+    function: ' CV^ORMARKER',
+    domain: 'PATIENT',
+    description: 'Identifies Combat Vet Marker',
+    rpcType: 'READ',
+    complexity: 'LOW',
+    returnType: 'ARRAY',
+    files: ['2'],
+    parameters: [{
```
mfuroyama: Ooh, also, can we rename this to `inputParameters`? This is important because the RPC emulation dispatcher actually uses the fields in `inputParameters` to do pre-process validation.

FYI: the `parameters` array is used to list any "Parameter Service" types used. In the case of the Patient RPCs, generally speaking, this can be left empty.

emulation/models/or-get-combat-vet.js
```diff
@@ -0,0 +1,48 @@
...
+        id: 'DFN', // in URN
+        parameterType: 'LITERAL',
+        required: true,
+        description: 'Patient URN',
+        example: 'urn:vista:999:2:25',
+    }],
+    labels: ['META'],
+    services: ['patientMetaService.lookupPatientCombatVet'],
+    inputParameters: [],
```
mfuroyama: Yeah, we can rename this to `parameters`, which should be empty for this RPC.

PR #225, opened 2018-02-09
vam-529-orwpt-id-info

pcchanghawaii:
this replaces vam-367-orwpt-id-info based on review comments 
https://github.com/vistadataproject/VICSServer/pull/214
vam 367 was created before the Patient collection was loaded and before the new VICSServer tag.

This branch is based on vam-502-or-get-combat-vet  so all the combat files are included as well.


emulation/models/or-get-combat-vet.js
```diff
@@ -0,0 +1,48 @@
...
+                    ret = 'CV';
+                }
+            }
+        }
+
+        return [ret];
+    },
+}];
+
+module.exports.rpcEmulatorModel = RPC_EMULATOR_MODEL;
```
mfuroyama: Gonna skip these files, but see the comments I made in pull request #224

emulation/models/orwpt-id-info.js
```diff
@@ -0,0 +1,52 @@
...
+            required: true,
+            description: 'Patient URN',
+            example: 'urn:vista:999:2:20',
+        }],
+        labels: ['META', 'hardcoded'],
+        services: ['patientMetaService.lookupPatientInfo'],
+        inputParameters: [],
+        invokeRPC: (args, session, callback) => {
+            patientMetaService.lookupPatientInfo(args, session, callback);
+        },
```
mfuroyama: Yeah, for the most part, it's just modifying stuff in the RPC model for convention sake:
5. The call to `lookupPatientInfo` should pass in the converted patient ID URN instead of the `args` array, since the service function accepts a patient Id

PR #226, opened 2018-02-10
vam-372-orwpt-diedon

pcchanghawaii:
https://github.com/vistadataproject/VICSServer/tree/vam-529-orwpt-id-info
based on the vam-529-orwpt-id-info and add the orwpt diedon implementation, particularly,  the following files.

emulation/models/orwpt-diedon.js
emulation/models/tests/orwpt-diedon-spec.js
services/patientMetaService/getPatientDiedOn.js
services/tests/getPatientDiedOn-spec.js


emulation/models/orwpt-id-info.js
```diff
@@ -0,0 +1,52 @@
...
+            required: true,
+            description: 'Patient URN',
+            example: 'urn:vista:999:2:20',
+        }],
+        labels: ['META', 'hardcoded'],
+        services: ['patientMetaService.lookupPatientInfo'],
+        inputParameters: [],
+        invokeRPC: (args, session, callback) => {
+            patientMetaService.lookupPatientInfo(args, session, callback);
+        },
```
mfuroyama: Yeah, for the most part, it's just modifying stuff in the RPC model for convention sake:
5. The call to `lookupPatientInfo` should pass in the converted patient ID URN instead of the `args` array, since the service function accepts a patient Id

emulation/models/orwpt-diedon.js
```diff
@@ -0,0 +1,41 @@
...
+        required: true,
+        description: 'Patient URN',
+        example: 'urn:vista:999:2:19',
+    }],
+    labels: ['META'],
+    services: ['patientMetaService.getPatientDiedOn'],
+    inputParameters: [],
+    invokeRPC: (args, session, callback) => {
+        patientMetaService.getPatientDiedOn(args, session, callback);
+    },
```
mfuroyama: Yeah, for the most part, it's just modifying stuff in the RPC model for convention sake:
5. The call to `getPatientDiedOn ` should pass in the converted patient ID URN instead of the `args` array, since the service function accepts a patient Id

PR #227, opened 2018-02-10
vam-382-orvaa-vaa

pcchanghawaii:
orvaa vaa implementation. I use INSURANCE TYPE (2_312) in file 2 for policy name lookup. We need to import file 2.312 and 36 if insurance type is the correct field to look up VAA. There is no appropriate patient data with insurance type value in Mongo to verify the expected result/behavior.  I may need to convert the type pointer to name or vise versa once the insurance type value is populated into Mongo.  

This branch is based on vam-529-orwpt-id-info
Newly added files are
emulation/models/orvaa-vaa.js
emulation/models/tests/orvaa-vaa-spec.js
services/tests/getPatientVaa-spec.js
services/patientMetaService/getPatientVaa.js


emulation/models/orvaa-vaa.js
```diff
@@ -0,0 +1,48 @@
...
+        required: true,
+        description: 'Patient URN',
+        example: 'urn:vista:999:2:19',
+    }],
+    labels: ['META'],
+    services: ['patientMetaService.getPatientVaa'],
+    inputParameters: [],
+    invokeRPC: (args, session, callback) => {
+        patientMetaService.getPatientVaa(args, session, callback);
+    },
```
mfuroyama: Yeah, for the most part, it's just modifying stuff in the RPC model for convention sake:
5. The call to `getPatientVaa ` should pass in the converted patient ID URN instead of the `args` array, since the service function accepts a patient Id

PR #228, opened 2018-02-12
Vam-530-orwpt-list-all

pcchanghawaii:
This branch is based on vam-382-orvaa-vaa-new
Newly added files are
emulation/models/orwpt-list-all.js
emulation/models/tests/orwpt-list-all-spec.js
services/tests/listPatientNames-spec.js
services/patientMetaService/listPatientNames.js


services/patientMetaService/listPatientNames.js
```diff
@@ -0,0 +1,35 @@
...
+function listPatientNames(context, callback) {
+    /* List all patient names */
+    this.find('Patient', {
+        query: { localStationNumber: context.stationId },
+        projection: serviceUtils.createProjection(['id', 'name']),
+    })
+        .then((elementArr) => {
+            const elementLabel = element => element.name;
+            const element = el => ({ id: serviceUtils.getIdFromUrn(el.id), name: el.name });
+            const res = _(elementArr).sortBy(elementLabel).map(element).value();
```
mfuroyama: You don't really need the `elementLabel` function here; if you're sorting an object array by a specific field in the objects, lodash lets you specify the name of the field as a string as the argument to `_.sortBy`:
```Javascript
const res = _(elementArr).sortBy('name').map(element).value();
```

PR #230, opened 2018-02-13
Vam 506 orqqal list

kekuso:
created ORQQAL LIST model and test specs


emulation/models/orqqal-list.js
```diff
@@ -0,0 +1,45 @@
...
+        services: ['allergyService.listAllergies'],
+        inputParameters: [{
+            id: 'PATIENT_ID',
+            parameterType: 'LITERAL',
+            required: false,
+            description: 'patient ID, not utilized in this implementation',
+            example: '25',
+        }],
+        invokeRPC: (args, session, callback) => {
+            allergyService.listAllergies(session, callback);
```
mfuroyama: Hmm, for now, let's add a check here that compares the patient IEN parameter value passed in against what's in the session object.

PR #231, opened 2018-02-13
vam-531-orwpt-select

pcchanghawaii:
This branch is based on  vam-502-or-get-combat-vet
Newly added files are 

emulation/models/orwpt-select.js
emulation/models/tests/orwpt-select-spec.js
services/patientMetaService/getSelectedPatientInfo.js
services/tests/getSelectedPatientInfo-spec.js


emulation/models/orwpt-select.js
```diff
@@ -0,0 +1,170 @@
...
+     * NAME^SEX^DOB^SSN^LOCIEN^LOCNM^RMBD^CWAD^SENSITIVE^ADMITTED^CONV^SC^SC%^
+     * 14 15 16 17ICN^AGE^TS^TSSVC
+        //CARTER,DAVID^M^2810302^000000113^^^^^0^^0^0^^^36^0 CARTER,DAVID
+        //0 sensitive 0 conv 0 SC 0 TS
+        //TWENTY,PATIENT T^M^2570102^411555432^^^^^1^^0^0^^^61^0 TWENTY,PATIENT T
+        //1 sensitive  61 age 0 SC 0 TS
+     */
+    formatResult: res => {
+        let ret = '';
+        if (res[0]) {
```
mfuroyama: Not a big deal, but we might want to consider checking for the negative case (`if (!res[0])`) and returning early on error. You can get rid of a level of indentation that way, which makes for slightly more readable code.

emulation/models/orwpt-select.js
```diff
@@ -0,0 +1,170 @@
...
+                ret += `^${res[0].ts}`;
+            } else {
+                ret += '^0 ';
+            }
+
+            if (res[0].name) {
+                ret += res[0].name;
+            }
+        }
+        return ret;
```
mfuroyama: This is cool, it works, but in these cases, what I've been telling the gang here in Hawaii to do in these caret-delimited MUMPS-y style formatting deals is to push all the output values to an array of strings, then call `join('^');` on the array to generate the final result string. So like: 

```Javascript
const result = [];
...
result.push(res[0].dateOfBirth || '');
result.push(res[0].socialSecurityNumber || '');
result.push(res[0].divisionId || '');
...
return result.join('^');
```
Either way though...just a suggestion. 😄

PR #233, opened 2018-02-16
VAM-558: Fix GMV PARAMETER getPar

creindle:


services/tests/gmvParameter-spec.js
```diff
@@ -52,11 +42,10 @@ describe('gmvParameter asynchronous specs', () => {
             testCase.args.push(context, (err, res) => {
                 if (err) return serviceTestUtils.handleError(err, done);
                 expect(res).toBeDefined();
-                expect(res).toEqual(jasmine.any(Object));
-
+                expect(res).toEqual(testCase.response);
                 // Remove some of the untestable keys
-                const modifiedResult = _.omit(res, testCase.omitKeys || []);
-                expect(modifiedResult).toEqual(testCase.response);
+                // const modifiedResult = _.omit(res, testCase.omitKeys || []);
```
creindle: I wasn't too sure what the intent of this was, I was thinking of removing it. Should I remove it?

services/tests/gmvParameter-spec.js
```diff
@@ -20,30 +20,20 @@ describe('gmvParameter asynchronous specs', () => {
         response: { srv: '3rd arg must include \'GMV\'', key: 'no_vital' },
     }, {
         name: 'entVal args',
-        args: ['ENTVAL', '1;LOC(44,', 'GMV DEFAULT VALUES', null, null],
+        args: ['ENTVAL', '1;LOC(44,', 'GMV DLL VERSION', null, null],
```
creindle: All of this was changed to test the "YES"/"NO" value.

emulation/models/gmv-parameter.js
```diff
@@ -74,8 +74,6 @@ const RPC_EMULATOR_MODEL = [
         },
 
         formatResult: (res) => {
-            const outputArr = ['', ''];
```
mfuroyama: Let's put this back the way it was

services/cprsMetaService/gmvParameter.js
```diff
@@ -119,18 +120,22 @@ function gmvParameter(routine, entArg, gmvPrmArg, instArg, valArg, context, call
...
     function GETPARRoutine(cb) {
-        prmService.get(gmvPrmArg, { entity: entArg, instance: instArg }, cb);
+        prmService.get(gmvPrmArg, { entity: entArg, instance: instArgSet }, cb);
     }
     function GETPARProcess(res, cb) {
-        cb(null, res[0] || {});
+        const yesNoParameters = ['GMV DLL VERSION', 'GMV GUI VERSION', 'GMV ALLOW USER TEMPLATES'];
+        const resValue = _.get(res[0], 'value', '');
+        if (yesNoParameters.includes(gmvPrmArg) && resValue > 0) return cb(serviceUtils.bail({ value: 'YES' }));
+        else if (yesNoParameters.includes(gmvPrmArg) && resValue <= 0) return cb(serviceUtils.bail({ value: 'NO' }));
```
mfuroyama: Consider replacing these two lines with something a little more DRY:
```Javascript
if (yesNoParameters.includes(gmvPrmArg)) {
    const value = (resValue > 0) ? 'YES' : 'NO';
    return cb(serviceUtils.bail({ value });
}
```

services/cprsMetaService/gmvParameter.js
```diff
@@ -21,6 +21,7 @@ const VISTAService = require('../VISTAMetaService');
 
 function gmvParameter(routine, entArg, gmvPrmArg, instArg, valArg, context, callback) {
     const service = this;
+    const instArgSet = instArg || 1;
```
mfuroyama: Yes, perfect!

emulation/models/gmv-parameter.js
creindle: My mistake, but glad we talked about this! It made so much more sense!

services/tests/gmvParameter-spec.js
mfuroyama: It looks like the code here is removing certain fields out of the result object that aren't deterministic (e.g. `id`) and then comparing the whole result object to the test case result. Without this, it looks like the tests will fail

mfuroyama: Oh, never mind. This is fine, since you modified the value returned by the GETPAR sub-function

mfuroyama:
So yeah, like we talked about, let's add more test cases to the service function Jasmine spec.

services/cprsMetaService/gmvParameter.js
```diff
@@ -70,13 +72,13 @@ function gmvParameter(routine, entArg, gmvPrmArg, instArg, valArg, context, call
 
     // 'ENTVAL' load domain info via entity value
     function ENTVALRoutine(cb) {
-        if (entArg === 'USR') return cb(serviceUtils.bail(context.name));
+        if (entArg === 'USR') return cb(serviceUtils.bail({ label: context.userLabel }));
         else if (entArg === 'DIV') {
-            return cb(serviceUtils.bail({ srv: context.division.label, key: context.division.label }));
+            return cb(serviceUtils.bail({ label: context.divisionLabel }));
```
creindle: The `sessionContext.json` has updated the key for this.

services/cprsMetaService/gmvParameter.js
```diff
@@ -70,13 +72,13 @@ function gmvParameter(routine, entArg, gmvPrmArg, instArg, valArg, context, call
...
         } else if (entArg === 'SYS') {
             return service.find('KernelSystemParameters', {
                 query: { localStationNumber: context.stationId },
-                projection: serviceUtils.createProject(['domainName.label']),
+                projection: serviceUtils.createProjection(['domainName.label']),
```
creindle: Luckily, you told me to run many more tests because this was actually originally incorrect!! Good catch!

services/cprsMetaService/gmvParameter.js
```diff
@@ -91,7 +93,7 @@ function gmvParameter(routine, entArg, gmvPrmArg, instArg, valArg, context, call
                 projection: serviceUtils.createProjection(['label']),
             }, cb);
         }
-        return cb(serviceUtils.bail(new Error('Incorrect entArg')));
+        return cb(serviceUtils.bail(response.no_ent));
```
creindle: This would better match for the `formatResult` portion in the model.

services/cprsMetaService/gmvParameter.js
```diff
@@ -103,19 +105,31 @@ function gmvParameter(routine, entArg, gmvPrmArg, instArg, valArg, context, call
     }
 
     // 'GETLST' retrieves all content of chosen parameter
-    function GETLISTRoutine(cb) {
-        if (!entArg) return cb(serviceUtils.bail(response.no_ent));
+    function GETLSTRoutine(cb) {
```
creindle: Ensuring the correct name of the routine matches with the function name. `GETLST` is the correct spelling.

services/cprsMetaService/gmvParameter.js
```diff
@@ -147,7 +161,7 @@ function gmvParameter(routine, entArg, gmvPrmArg, instArg, valArg, context, call
     const queryRoutine = {
         DELPAR: [DELPARRoutine, DELPARRemove, DELPARProcess],
         ENTVAL: [ENTVALRoutine, ENTVALProcess],
-        GETLIST: [GETLISTRoutine, GETLISTProcess, GETLISTQueryResponse],
+        GETLST: [GETLSTRoutine, GETLSTProcess, GETLSTQueryResponse],
```
creindle: Luckily, that caught the mistakes here! Since `GETLST` was not even being called in the service spec file, we would've never known we weren't calling this at all!

services/cprsMetaService/gmvParameter.js
```diff
@@ -51,9 +51,11 @@ function gmvParameter(routine, entArg, gmvPrmArg, instArg, valArg, context, call
     // if GMV args does not start with letters "GMV", return error str
     if (_.isString(gmvPrmArg) && !gmvPrmArg.match(/^GMV.*/)) return callback(null, response.no_vital);
 
+    // if there is no entity argument, return error str
+    if (!entArg) return callback(null, response.no_ent);
```
creindle: This code should've been used in many places, so it was best to check at top level.

PR #234, opened 2018-02-16
Vam 540 implement write functions for allergy service

kekuso:
Implemented Create, remove, unremove, and delete services for Allergies


services/allergyService/createAllergy.js
```diff
@@ -0,0 +1,150 @@
...
+    // check for required input properties
+    let requiredProp = '';
+    const validInput = REQUIRED_INPUT_PROPERTIES.every((property) => {
+        if (!Object.keys(args).includes(property)) {
+            requiredProp = property;
+            return false;
+        }
+        return true;
+    });
+    if (!validInput) return callback(null, { failure: `Failed to create Allergy (${requiredProp} required)` });
```
mfuroyama: 👍 This is great, so good in fact that we might want to utility-ize this at some point...

PR #235, opened 2018-02-18
Vam-371 dg-sensitive-record-access

pcchanghawaii:
Remarks for this file:
services/patientMetaService/lookupDgSensitiveRecord.js   
* is not eslint compliant yet but pardon me for debugging's purpose until the merge..
* my jasmine doesn't recognize spread operator "..."  yet so I left the apply function
* MVDM files under the script folder must be added in order to pass the test spec (lookupDgSensitiveRecord-spec.js).
Those additional json entries are used to create conditional data to meet the test spec.


services/patientMetaService/lookupDgSensitiveRecord.js
```diff
@@ -0,0 +1,465 @@
...
+        }
+        cb(null, patArg[0]);
+    }
+
+    function lookupSensitivity(pat, cb) {
+        service.find('DgSecurityLog', {
+            query: {
+                patientName: {
+                    id: pat.id,
+                    label: pat.name,
```
mfuroyama: You probably don't need the extra `label` in the query. The `id` should be unique enough to give you what you want, I'm guessing?

services/patientMetaService/lookupDgSensitiveRecord.js
```diff
@@ -0,0 +1,465 @@
...
+                    }
+                }
+            });
+        }
+        cb(null);
+    }
+
+    function getMasRestrict(cb) {
+        service.find('MasParameters', {
+            query: { localStationNumber: 999 }, // context.stationId}, // id: '43-1' need to change it to select all
```
mfuroyama: Eventually, before we move the code in, we'll need to remove this hard-coded station ID...

services/patientMetaService/lookupDgSensitiveRecord.js
```diff
@@ -0,0 +1,465 @@
...
+            patientSensitive === 1 &&
+            (masRestrict | dgSensitiveKey | patientIsEmployee | isPatientUser()) === 0) {
+            // console.log('case 1...');
+            sensitiveAccessResult = 2;
+            arrayResult.push(sensitiveAccessResult);
+            arrayResult.push.apply(arrayResult, warningAcknowledge);
+        } // 1
+        else if (userSsn &&
+            (patientSensitive & dgSensitiveKey) === 1 &&
+            (masRestrict | patientIsEmployee | isPatientUser) === 0) {
```
mfuroyama: Hmm, from this point on, it looks like you're including the actual **function reference** in your logical comparison, as opposed to the **function result**. I'm guessing this isn't what you intended?

Also, on a related note, since the result of `isPatientUser()` never changes, one thing to consider is calling `isPatientUser()` once and using that result in the logical comparisons.

services/patientMetaService/lookupDgSensitiveRecord.js
pcchanghawaii: Supposedly, but if I take out the label, then the query won't return any entry even if the id exists. Maybe there's something under the hook. I will leave the label for now.

mfuroyama: Ahh, okay...sounds good!

PR #241, opened 2018-02-22
VAM-508: ORQQPL CHECK DUP

creindle:
[X] Removed `console.log`s from `listClinicalProblem.js` and `listClinicalProblem-spec.js` (which is why there are more files than usual)


emulation/models/orqqpl-check-dup.js
```diff
@@ -0,0 +1,56 @@
...
+        description: 'Check for duplicate problem associated with a problem IEN.',
+        rpcType: 'READ',
+        complexity: 'LOW',
+        returnType: 'ARRAY',
+        files: ['9000011', '9000011.11'],
+        parameters: [],
+        labels: ['JLV', 'RESULTS-UNCHECKED'],
+        services: ['problemService.checkDuplicateEntry'],
+        inputParameters: [{
+            id: '',
```
creindle: Wasn't 100% sure how to fill out this info for the activeFilter inputParam. Is this id: 'DFN', parameterType: 'LITERAL'?

emulation/models/orqqpl-check-dup.js
```diff
@@ -0,0 +1,56 @@
...
+            example: 'R',
+        }, {
+            id: 'PIFN',
+            parameterType: '',
+            required: true,
+            description: 'the problem IEN',
+            example: '1',
+        }, {
+            id: '',
+            parameterType: '',
```
creindle: I believe the type if "LITERAL" and I saw required was "TRUE" in the native VISTA RPC doc. Should I change that here?

emulation/models/orqqpl-check-dup.js
mfuroyama: Exactly! Use what you got...

emulation/models/orqqpl-check-dup.js
mfuroyama: Yeah:
```Javascript
    parameterType: 'LITERAL',
    required: false,
```
👌

services/problemService/checkDuplicateEntry.js
```diff
@@ -0,0 +1,47 @@
...
+
+/*
+* Check for duplicate input based on matching existing problem object URN to input problem URN
+* @param {String} resURN Problem object URN from list of problems of user
+* @param {String} problemURN Problem URN to search for
+* @return {(Number|String)} 0 if duplicate problem not found or id of problemURN if duplicate is found
+*/
+function checkProblemDuplicateEntry(resURN, problemURN) {
+    if (resURN !== problemURN) return 0;
+    return serviceUtils.getIdFromUrn(resURN);
```
mfuroyama: Let's return the URN of the duplicate problem here instead of the IEN, and have the model `formatResult` function handle the URN-to-IEN conversion.

Remember, service functions should only deal with URNs!

services/problemService/checkDuplicateEntry.js
```diff
@@ -0,0 +1,47 @@
...
+ * @memberof problemService
+ * @instance
+ */
+function checkDuplicateEntry(activeFilter, problemURN, context, callback) {
+    const service = this;
+    function problemQuery(cb) {
+        service.listClinicalProblem(activeFilter, context, cb);
+    }
+    function processDuplicateEntry(res, cb) {
+        const resURN = _.get(res[0], 'id', '');
```
mfuroyama: Remember that the point of this function is to see if the list of problems associated with a patient already contains a problem with the id of `problemURN`.

So, instead of just looking at the first value of the list, a more valid approach would be to do something like a `find` on the list, then return the URN if there's a match and `0` if not

emulation/models/orqqpl-check-dup.js
```diff
@@ -0,0 +1,56 @@
...
+
+        invokeRPC: (rpcArgs, session, callback) => {
+            const acceptableFilters = ['R', 'A', 'I'];
+            const activeFilter = (_.includes(acceptableFilters, rpcArgs[0])) ? rpcArgs[0] : '';
+            const problemURN = modelUtils.createURNFromParams('9000011', rpcArgs[1], session.stationId);
+            problemService.checkDuplicateEntry(activeFilter, problemURN, session, callback);
+        },
+        formatResult: (res) => {
+            if (!res) return 0;
+            return `${res}^A`;
```
mfuroyama: As mentioned below, let's have the service function return a URN, and let's do the URN-to-IEN conversion here

services/tests/listClinicalProblem-spec.js
```diff
@@ -12,7 +12,6 @@ describe('listClinicalProblem asynchronous specs', () => {
 
     it('Returns list of all problems that match patientId in context object', (done) => {
         modelFunction.listClinicalProblem('R', context, (err, res) => {
-            console.log('didididi', res);
```
mfuroyama: LOL!! Thanks for catching this

services/problemService/checkDuplicateEntry.js
creindle: I think I understand this! I'll push up changes so you can see if I do. Thank you for explaining!

PR #242, opened 2018-02-23
Vam 514 orqqpl problem list

sidjtd:
2 Files for orqqpl-problem-list


emulation/models/orqqpl-problem-list.js
```diff
@@ -0,0 +1,109 @@
...
+            },
+            {
+                id: 'GMPSTAT',
+                parameterType: 'LITERAL',
+                description: 'Problem status filter: A - active, I - inactive, B - Both active and inactive, R - removed',
+                example: 'A', // ACTIVE
+            },
+        ],
+        invokeRPC: (rpcArgs, session, callback) => {
+            // rpcArgs[0] is 'DFN' or Patient ID and not to be used to avoid other users searching other user lists
```
mfuroyama: Later on, we'll add security measures here to ensure that the `DFN` corresponds to the value in the `session.patientId` value. This adds a layer of security and ensures that users only look at patient info that they're allowed to (determined through the 'patient select' process)

For now though, this is fine as is

emulation/models/orqqpl-problem-list.js
```diff
@@ -0,0 +1,109 @@
...
+                required: true,
+                description: 'Record number of the patient from the Patient File',
+                example: '62',
+                // Will not utilized in this service Function, only allow user signed in to view their own list
+            },
+            {
+                id: 'GMPSTAT',
+                parameterType: 'LITERAL',
+                description: 'Problem status filter: A - active, I - inactive, B - Both active and inactive, R - removed',
+                example: 'A', // ACTIVE
```
mfuroyama: Shoot...I missed this the first pass, let's add `required: true` here to ensure the parameter is there for `invokeRPC`

PR #244, opened 2018-02-23
VAM-499: GMV LATEST VM 

creindle:
I realized that there was another comment to format the results in a particular order. I'll be working on that tomorrow since that's the last thing I need to do.


emulation/models/gmv-latest-vm.js
```diff
@@ -0,0 +1,105 @@
...
+            let vitalTakenDateTime = _.get(vital, 'vitalsTakenDateTime', '');
+            vitalTakenDateTime = moment(vitalTakenDateTime, 'YYYY-MM-DDTHH:mm:ss').format('MM/DD/YY@HH:mm');
+
+            output.push(`${setVitalType[vitalType]}:    `);
+            output.push(`(${vitalTakenDateTime})`);
+            if (vitalType === 'PULSE OXIMETRY') output.push(`${vitalValue}${vitalUnits}`);
+            if (vitalType === 'HEIGHT') output.push(`${convertInchToFeet(vitalValue)}`);
+            if (vitalType !== 'PULSE OXIMETRY' && vitalType !== 'HEIGHT') output.push(vitalValue);
+            if (metricValue) {
+                if (vitalType !== 'HEIGHT') output.push(vitalUnits);
```
creindle: There must be a better way to program this. Let me know if I was way to specific or if I was correct (I was trying to match the output in the JIRA comments)

emulation/models/gmv-latest-vm.js
```diff
@@ -0,0 +1,105 @@
...
+            const vitalUnits = _.get(vital, 'units', '');
+            const metricValue = _.get(vital, 'metricValue', '');
+            const metricUnits = _.get(vital, 'metricUnits', '');
+            const enteredBy = _.get(vital, 'enteredBy.label', '');
+            let vitalTakenDateTime = _.get(vital, 'vitalsTakenDateTime', '');
+            vitalTakenDateTime = moment(vitalTakenDateTime, 'YYYY-MM-DDTHH:mm:ss').format('MM/DD/YY@HH:mm');
+
+            output.push(`${setVitalType[vitalType]}:    `);
+            output.push(`(${vitalTakenDateTime})`);
+            if (vitalType === 'PULSE OXIMETRY') output.push(`${vitalValue}${vitalUnits}`);
```
creindle: Was I correct in this? I saw in the output that this was the only vital type that had the unit right next to the value.

emulation/models/gmv-latest-vm.js
```diff
@@ -0,0 +1,107 @@
...
+            const vitalType = _.get(vital, 'vitalType.label', '');
+            const vitalValue = _.get(vital, 'value', '');
+            const vitalQualifier = _.get(vital, 'qualifier', []);
+            const vitalUnits = _.get(vital, 'units', '');
+            const metricValue = _.get(vital, 'metricValue', '');
+            const metricUnits = _.get(vital, 'metricUnits', '');
+            const enteredBy = _.get(vital, 'enteredBy.label', '');
+            let vitalTakenDateTime = _.get(vital, 'vitalsTakenDateTime', '');
+            vitalTakenDateTime = moment(vitalTakenDateTime, 'YYYY-MM-DDTHH:mm:ss').format('MM/DD/YY@HH:mm');
+            output.push(`${setVitalType[vitalType]}:    `);
```
creindle: I don't like how I had specifically 4 spaces, what is a better way to add a tab?

emulation/models/gmv-latest-vm.js
```diff
@@ -0,0 +1,107 @@
...
+            'BLOOD PRESSURE': 'B/P',
+            TEMPERATURE: 'Temp.',
+            RESPIRATION: 'Resp.',
+            PULSE: 'Pulse',
+            HEIGHT: 'Ht.',
+            WEIGHT: 'Wt.',
+            'CENTRAL VENOUS PRESSURE': 'CVP',
+            'PULSE OXIMETRY': 'Pulse Ox',
+            CIRCUMFERENCE_GIRTH: 'Circ/Girth',
+            'BODY MASS INDEX': 'Body Mass Index',
```
creindle: I wasn't too sure what this label would be as I didn't see the data in the DB, so this is what I set it as.

mfuroyama: This is actually calculated and will need to be added later into ORQQVI VITALS and into this and other Vitals RPCs, but for now it's fine

emulation/models/gmv-latest-vm.js
mfuroyama: If you want to duplicate tab-type spacing, might want to see if the lodash `tabEnd` function works for you.

emulation/models/tests/gmv-latest-vm-spec.js
```diff
@@ -0,0 +1,323 @@
...
+        'Pain:      (02/01/17@16:06)  7  _ALEXANDER,ROBERT_Vitals',
+    ];
+
+    it('invoking formatResult with a valid array should return specific output', () => {
+        const res = rpc.formatResult(TEST_INPUT);
+        expect(res).toBeDefined();
+        expect(res).toEqual(jasmine.any(Array));
+        expect(res.length).toEqual(10);
+        expect(res).toEqual(TEST_OUTPUT);
+    });
```
mfuroyama: Thank you for creating very thorough test cases!!

emulation/models/gmv-latest-vm.js
mfuroyama: This is fine for now!

PR #247, opened 2018-02-23
vam-532-orwpt-top

pcchanghawaii:
rely on patientId from session. assume return empty '' if no patient selected yet.


mfuroyama:
Hey @pcchanghawaii, let's put this on the back-burner for a little bit. It looks like the `ORWPT TOP` results depends on values set in the `ORWPT SELECT` RPC. Specifically, we'll have to mirror the MUMPS set to `^DISV(DUZ,"^DPT(")` in `ORWPT SELECT` in the MongoDB code somehow, since `ORWPT TOP` uses that global value in the MUMPS to return its result.

Anyway, I created a [follow-up issue](https://hawaiirg.atlassian.net/browse/VAM-563) to track this work and I'm tagging @conorcaregraf to keep him in the loop. Thanks!

PR #248, opened 2018-02-26
vam-374-orwpt1-prcare

pcchanghawaii:
OUTPATIENT PROFILE (404_41), we need this file.  I'm using those fields from Patient to extract the same info.


emulation/models/orwpt1-prcare.js
```diff
@@ -0,0 +1,48 @@
...
+        required: true,
+        description: 'Patient IEN',
+        example: '19',
+    }],
+    labels: ['META', 'hardcoded', 'RESULTS-UNCHECKED'],
+    services: ['patientMetaService.getPrimaryCare'],
+    parameters: [],
+    invokeRPC: (args, session, callback) => {
+        const urnObj = {
+            fileType: 2, id: args[0], isNational: false, station: '999',
```
mfuroyama: We should probably remove the hard-coding of `station` here and set it to the `session.stationId` value.

emulation/models/orwpt1-prcare.js
```diff
@@ -0,0 +1,48 @@
...
+        patientMetaService.getLastSelected(urnStr, session, callback);
+    },
+
+    formatResult: (res) => {
+        const ret = [];
+        ret.push(_.get(res[0], 'currentPcTeam.label', '')); // id or label??
+        ret.push(_.get(res[0], 'currentPcPractitioner.label', ''));
+        ret.push(_.get(res[0], 'attendingPhysician.label', ''));
+        ret.push(_.get(res[0], 'attendingPosition.label', ''));
+        ret.push(_.get(res[0], 'provider.label', ''));
```
mfuroyama: Actually, looking at the MUMPS, I think all of these are IDs, which means the IEN values need to be teased out of the MongoDB result `id` fields

PR #250, opened 2018-02-28
VAM-564, VAM-565: Followup ORQQVI VITALS and GMV LATEST VM

creindle:
[X] ORQQVI VITALS passes all tests in extra spec file
[X] GMV LATEST VM passes all tests in extra spec file
[X] Renamed `listVitals` instances to `lookupLatest` to reflect intention of service function
[X] `unitConversionUtils` file created to contain all vitals function conversions
[X] `convertFMtoISO` file edited to check edge cases


creindle:
I realized, I can merge the GMV LATEST VM fix with this PR!

services/vitalsService/unitConversionUtils.js
```diff
@@ -0,0 +1,43 @@
...
+    const remainder = inches % 12;
+    return `${feet} ft ${remainder} in`;
+};
+
+/*
+* @summary Calculates metricValue from Vital collection 'CENTRAL VENOUS PRESSURE' centimeters value of water to millimeters of mercury
+* @param {Number} cmOfWater Centimeters of water
+* @return {Number} Converted millimeters of mercury
+*/
+const calculateH2OToM = cmOfWater => cmOfWater * 0.73553888590369;
```
creindle: I realized I needed to add a `.toFixed(2)` to the end of this. Either `(1)` or `(2)`

emulation/models/gmv-latest-vm.js
```diff
@@ -89,7 +84,7 @@ const RPC_EMULATOR_MODEL = [{
             output.push(_.padEnd(`${setVitalType[vitalType]}:`, 11));
             output.push(`(${vitalTakenDateTime})`);
             if (vitalType === 'PULSE OXIMETRY') output.push(`${vitalValue}${vitalUnits}`);
-            if (vitalType === 'HEIGHT') output.push(`${convertInchToFeet(vitalValue)}`);
+            if (vitalType === 'HEIGHT') output.push(`${vitalsService.calculateInchesToFeet(vitalValue)}`);
```
mfuroyama: Can we change the utility function name from `calculateInchesToFeet` to `convertInchesToFeet`? From an API naming perspective, it's cleaner

emulation/models/tests/orqqvi-vitals-extra-spec.js
```diff
@@ -162,9 +146,9 @@ xdescribe('ORQQVI VITALS', () => {
 
                 // BMI is last and 'ien' is -1
                 if (lineInfo.type === 'BMI') {
-                    expect(i).toEqual(res.result.length - 1); // last one
+                    expect(i).toEqual(res.length - 1); // last one
```
mfuroyama: Thanks for fixing that

services/vitalsService/lookupLatest.js
```diff
@@ -14,32 +15,55 @@ const moment = require('moment');
...
     const momentEndDate = endDate ? moment(endDate, moment.ISO_8601) : moment();
-
-    this.find('Vital', {
-        query: { 'patient.id': context.patientId, localStationNumber: context.stationId },
+    service.find('Vital', {
+        query: { 'patient.id': context.patientId, localStationNumber: context.stationId, isRemoved: { $ne: true } },
     }).then((res) => {
-        const inputtedVitals = res.filter((vital) => {
+        if (!res || _.isEmpty(res)) return callback(null, []);
+        const vitalsWithinDateRange = res.filter((vital) => {
```
mfuroyama: Consider doing this in the MongoDB query using the `$gte` and `$lt` criteria fields. Not only is this faster, but it makes the code cleaner.

mfuroyama: Also, consider sorting the results in MongoDB by adding an `options` object to the `find` options object:
```Javascript
service.find('Vital', {
    query: { 'patient.id': context.patientId, localStationNumber: context.stationId, isRemoved: { $ne: true } },
    options: { sort: { vitalsTakenDateTime: -1 } },
}
```

services/vitalsService/lookupLatest.js
```diff
@@ -14,32 +15,55 @@ const moment = require('moment');
...
+        if (!vitalsWithinDateRange || _.isEmpty(vitalsWithinDateRange)) return callback(null, []);
         // _.sortBy() sorts the array earliest to latest, but we want it by latest first
-        const dateSortedVitals = _.sortBy(inputtedVitals, 'vitalsTakenDateTime').reverse();
+        const dateSortedVitals = _.sortBy(vitalsWithinDateRange, 'vitalsTakenDateTime').reverse();
         // grab most recent
         const vitalsByType = _.groupBy(dateSortedVitals, 'vitalType.label');
         const mostRecentList = _.keys(vitalsByType).map(type => _.first(vitalsByType[type]));
+
+        // gather info to calculate H2O to Mercury for CVP label
+        const cvpIndex = _.findIndex(mostRecentList, ['vitalType.label', 'CENTRAL VENOUS PRESSURE']);
```
mfuroyama: Consider the case where your patient does not have a `'CENTRAL VENOUS PRESSURE'` vitals entry. Let's make sure that the code below doesn't crash if that's the case

services/vitalsService/unitConversionUtils.js
```diff
@@ -0,0 +1,43 @@
...
+    const weightCalc = parseFloat(weight) / 2.2;
+    return (weightCalc / (heightCalc * heightCalc)).toFixed(2);
+};
+
+/*
+* @summary Calculates inches to feets from Vital collection 'HEIGHT'
+* @param {Number} inches  Inches to convert
+* @return {String} Converted feet and inches
+*/
+const calculateInchesToFeet = (inches) => {
```
mfuroyama: Can we change this to `convert` instead of `calculate`? Thanks!

services/vitalsService/unitConversionUtils.js
mfuroyama: Same here: `calculate` => `convert`

services/dateTimeService/convertFMtoISO.js
```diff
@@ -17,10 +17,11 @@ const moment = require('moment');
  * @instance
  */
 function convertFMtoISO(fmDate) {
-    if (!_.isString(fmDate) || _.isNaN(_.toNumber(fmDate))) {
+    if (!_.isString(fmDate) && _.isNaN(_.toNumber(fmDate))) {
```
mfuroyama: Let's talk about this function. I want to know what your thought process was when you fixed this up.

services/vitalsService/lookupLatest.js
creindle: This was way better! Thank you for the suggestion!

services/dateTimeService/convertFMtoISO.js
```diff
@@ -17,10 +17,10 @@ const moment = require('moment');
...
  */
 function convertFMtoISO(fmDate) {
-    if (!_.isString(fmDate) || _.isNaN(_.toNumber(fmDate))) {
-        return null;
-    }
-    const parts = fmDate.split('.');
+    if (!fmDate) return null;
+    const strFmDate = _.toString(fmDate);
+    if (_.isNaN(_.toNumber(strFmDate))) return null;
+    const parts = strFmDate.split('.');
```
mfuroyama: 👍

services/vitalsService/lookupLatest.js
```diff
@@ -0,0 +1,70 @@
...
+ * @memberof vitalsService
+ * @instance
+ */
+
+function lookupLatest(startDate, endDate, context, callback) {
+    const service = this;
+    // date before records were collected
+    const momentStartDate = startDate ? moment(startDate) : moment('1900-01-01T00:00:00', moment.ISO_8601);
+    // current date time
+    const momentEndDate = endDate ? moment(endDate, moment.ISO_8601) : moment();
```
mfuroyama: So here, since we're not doing date comparison, we don't need to do the conversion to `moment` objects. What we **do** need to do, however, is adjust the query for efficiency and functionality. What I mean is, we **only** want to add the `$gte` or `$lt` criteria to the `query` object if the `startDate` and `endDate` values are valid. Otherwise, we might end up jamming up the query

So, consider removing the above code setting the `momentStartDate` and `momentEndDate` and adding some "query building" code instead, like so:
```Javascript
    const query = {};
    if (startDate) {
        _.set(query, 'vitalsTakenDateTime.$gte', startDate);
    }
    if (endDate) {
        _.set(query, 'vitalsTakenDateTime.$lt', endDate);
    }
```

Let me know if you need further clarification on this

emulation/models/gmv-latest-vm.js
mfuroyama: Don't forget to change the name on this function call

services/vitalsService/unitConversionUtils.js
```diff
@@ -0,0 +1,43 @@
...
+ * latest weight allowed in date range (seeing this would require very specific data)
+ *
+ * We are just taking latest height from date range for now along with latest weight.
+ *
+ * MUMPS CALC: S WT=WT/2.2, S HT=HT*2.54/100, Q $J(WT/(HT*HT),0,2)
+ */
+const calculateBMI = (weight, height) => {
+    const heightCalc = (parseFloat(height) * 2.54) / 100;
+    const weightCalc = parseFloat(weight) / 2.2;
+    return (weightCalc / (heightCalc * heightCalc)).toFixed(2);
```
mfuroyama: So, here's a case to consider: What if I (Mike) was building a Vitals RPC model, but *my* version of BMI needs show 3 decimal places of precision. In this case, I can't use your `calculateBMI` function, since it strips away the decimal place I need.

Consider making the utility function a little bit more flexible here: Instead of returning a string with a fixed 2-digit precision, you might want to:
1. Have the function return a floating-point number, and let the caller handle the precision
2. Add an optional 'digits' parameter to your function (defaulting to `2` here), to allow users to specify the precision of the returned value string.

The bigger "teachable moment" here is about creating utility functions. The main purpose of any utility function is to provide functionality in as wide a case as possible.

services/vitalsService/lookupLatest.js
creindle: Good point! Thank you for the example!

services/vitalsService/unitConversionUtils.js
creindle: This is a good point!!

emulation/models/gmv-latest-vm.js
```diff
@@ -87,10 +82,11 @@ const RPC_EMULATOR_MODEL = [{
             let vitalTakenDateTime = _.get(vital, 'vitalsTakenDateTime', '');
             vitalTakenDateTime = moment(vitalTakenDateTime, 'YYYY-MM-DDTHH:mm:ss').format('MM/DD/YY@HH:mm');
             output.push(_.padEnd(`${setVitalType[vitalType]}:`, 11));
-            output.push(`(${vitalTakenDateTime})`);
+            if (vitalType !== 'BODY MASS INDEX') output.push(`(${vitalTakenDateTime})`);
```
creindle: All of these if statements are so cluttered... I'm thinking of a way to cleaning it up...

emulation/models/gmv-latest-vm.js
```diff
@@ -87,10 +82,11 @@ const RPC_EMULATOR_MODEL = [{
             let vitalTakenDateTime = _.get(vital, 'vitalsTakenDateTime', '');
             vitalTakenDateTime = moment(vitalTakenDateTime, 'YYYY-MM-DDTHH:mm:ss').format('MM/DD/YY@HH:mm');
```
mfuroyama: Also, don't forget that the values specified as `DATE` type in the database schema will be returned by the data access layer as `Date` objects.

This means that you can clean this up a little bit by converting the `Date` object to an ISO string:
```
vitalTakenDateTime = moment(vitalTakenDateTime.toISOString()).format('MM/DD/YY@HH:mm');
```

emulation/models/gmv-latest-vm.js
```diff
@@ -87,10 +82,11 @@ const RPC_EMULATOR_MODEL = [{
...
             output.push(_.padEnd(`${setVitalType[vitalType]}:`, 11));
-            output.push(`(${vitalTakenDateTime})`);
+            if (vitalType !== 'BODY MASS INDEX') output.push(`(${vitalTakenDateTime})`);
             if (vitalType === 'PULSE OXIMETRY') output.push(`${vitalValue}${vitalUnits}`);
-            if (vitalType === 'HEIGHT') output.push(`${convertInchToFeet(vitalValue)}`);
-            if (vitalType !== 'PULSE OXIMETRY' && vitalType !== 'HEIGHT') output.push(vitalValue);
+            else if (vitalType === 'HEIGHT') output.push(`${vitalsService.convertInchesToFeet(vitalValue)}`);
+            else if (vitalType === 'PAIN' && vitalValue === '0') output.push('0 - No pain');
```
mfuroyama: Ooh, almost missed this: there are other modifiers associated with pain levels:
```
    10: 'Worst imaginable pain',
    99: 'Unable to respond',
```

creindle: That helps so much!

emulation/models/gmv-latest-vm.js
creindle: NOoooooo I should've caught thisssssss....

creindle: But thank you for the catch!

creindle: Ah I remember now! I caught this in the `orqqvi-vitals`!

emulation/models/gmv-latest-vm.js
```diff
@@ -85,12 +80,17 @@ const RPC_EMULATOR_MODEL = [{
...
-            output.push(`(${vitalTakenDateTime})`);
+            if (vitalType !== 'BODY MASS INDEX') output.push(`(${vitalTakenDateTime})`);
             if (vitalType === 'PULSE OXIMETRY') output.push(`${vitalValue}${vitalUnits}`);
-            if (vitalType === 'HEIGHT') output.push(`${convertInchToFeet(vitalValue)}`);
-            if (vitalType !== 'PULSE OXIMETRY' && vitalType !== 'HEIGHT') output.push(vitalValue);
+            else if (vitalType === 'HEIGHT') output.push(`${vitalsService.convertInchesToFeet(vitalValue)}`);
+            else if (vitalType === 'PAIN') {
+                if (vitalValue === '0') output.push('0 - No pain');
+                else if (vitalValue === '10') output.push('10: "Worst imaginable pain"');
+                else if (vitalValue === '99') output.push('99: "Unable to respond"');
```
mfuroyama: Ahh, shoot, sorry I wasn't clear about this. The string values here should be:
```Javascript
        else if (vitalValue === '10') output.push('10 - Worst imaginable pain');
        else if (vitalValue === '99') output.push('99 - Unable to respond');
```

emulation/models/gmv-latest-vm.js
```diff
@@ -85,12 +80,17 @@ const RPC_EMULATOR_MODEL = [{
             const metricUnits = _.get(vital, 'metricUnits', '');
             const enteredBy = _.get(vital, 'enteredBy.label', '');
             let vitalTakenDateTime = _.get(vital, 'vitalsTakenDateTime', '');
-            vitalTakenDateTime = moment(vitalTakenDateTime, 'YYYY-MM-DDTHH:mm:ss').format('MM/DD/YY@HH:mm');
+            vitalTakenDateTime = moment(vitalTakenDateTime.toISOString()).format('MM/DD/YY@HH:mm');
```
mfuroyama: Sorry for missing this earlier, but the output string should retain the UTC values so, and we can do that through Moment.js. Let's change this to:
```Javascript
vitalTakenDateTime = moment(vitalTakenDateTime.toISOString()).utc().format('MM/DD/YY@HH:mm');
```

emulation/models/tests/gmv-latest-vm-spec.js
```diff
@@ -298,26 +299,33 @@ describe('GMV LATEST VM asynchronous specs', () => {
...
+        'Pulse:       (02/01/18@06:54)  70  _ALEXANDER,ROBERT_Vitals',
+        'Resp.:       (02/01/18@06:54)  13  _ALEXANDER,ROBERT_Vitals',
+        'Pulse Ox:    (02/01/18@06:54)  96%  _ALEXANDER,ROBERT_Vitals',
+        'B/P:         (02/01/18@06:54)  115/75  (R ARM,SITTING)  _ALEXANDER,ROBERT_Vitals',
+        'Ht.:         (02/01/18@06:54)  5 ft 10 in  (177.8 cm)  _ALEXANDER,ROBERT_Vitals',
+        'Wt.:         (02/01/18@06:54)  165  lb  (75 kg)  _ALEXANDER,ROBERT',
+        'Body Mass Index:  23.72  __Vitals',
+        'CVP:         (02/01/18@06:54)  4  _ALEXANDER,ROBERT_Vitals',
+        'Circ/Girth:  (02/01/18@06:54)  32  in  (81.28 cm)  _ALEXANDER,ROBERT_Vitals',
+        'Pain:        (02/01/17@06:06)  7  _ALEXANDER,ROBERT_Vitals',
```
mfuroyama: Wait, these times should be expressed as UTC, so these should actually be `@16:54`. Let's change these back and modify the code as above

emulation/models/gmv-latest-vm.js
creindle: I... should've noticed this!

PR #253, opened 2018-03-02
VAM-497, VAM-566 - GMV CLOSEST READING Implementation and Extra Spec File Pass

creindle:
[X] Created model and service files
[X] All specs pass in `gmv-closest-reading-extra-spec` file with extra tests added
[X] Ensured `null` is returned in `dateTimeService.convertFMtoISO` service for invalid dates, specifically when the Fileman time is too large to convert.


emulation/models/gmv-closest-reading.js
```diff
@@ -0,0 +1,92 @@
...
+const RPC_EMULATOR_MODEL = [{
+    id: 'GMV_LATEST_VM',
+    name: 'GMV LATEST VM',
+    function: 'GETLAT^GMVGETD',
+    parameters: [
+        'GMVDFN',
+        'GMVDT',
+        'GMVT',
+        'GMVFLAG',
+    ],
```
mfuroyama: Let's remove the strings here. Remember that the `parameters` field is to document the names of any parameters we access through the Parameter service (e.g. `ORK SYSTEM ENABLE/DISABLE`)

emulation/models/gmv-closest-reading.js
```diff
@@ -0,0 +1,92 @@
...
+'use strict';
+
+const _ = require('lodash');
+const dateTimeService = require('../../services/dateTimeService');
+const vitalsService = require('../../services/vitalsService');
+
+const RPC_EMULATOR_MODEL = [{
+    id: 'GMV_LATEST_VM',
+    name: 'GMV LATEST VM',
+    function: 'GETLAT^GMVGETD',
```
mfuroyama: Let's change these to match the `GMV CLOSEST READING` RPC info

emulation/models/gmv-closest-reading.js
```diff
@@ -0,0 +1,92 @@
...
+    parameters: [
+        'GMVDFN',
+        'GMVDT',
+        'GMVT',
+        'GMVFLAG',
+    ],
+    labels: ['PATIENT'],
+    domain: 'VITALS',
+    files: ['120_5'],
+    services: ['vitalsService'],
```
mfuroyama: Let's change this to `'vitalsService.getClosestValue'`

emulation/models/tests/gmv-closest-reading-extra-spec.js
```diff
@@ -21,18 +21,18 @@
  */
 
 const unitTestUtils = require('./unitTestUtils');
-// const rpcEmulatorModel = require('../gmv-closest-reading').rpcEmulatorModel;
-const rpcEmulatorModel = [{}];
+const rpcEmulatorModel = require('../gmv-closest-reading').rpcEmulatorModel;
+// const rpcEmulatorModel = [{}];
```
mfuroyama: This was merely a placeholder, let's delete this line

emulation/models/tests/gmv-closest-reading-extra-spec.js
```diff
@@ -21,18 +21,18 @@
...
+describe('GMV CLOSEST READING', () => {
     let rpc;
     const patientMainVitalsIEN = '25';
 
     const patientMainVitalsURN = 'urn:vista:9998:2:25';
 
     const mainVitalsContext = {
         patientId: patientMainVitalsURN,
-        stationNumber: 9998,
+        stationId: 9998,
```
mfuroyama: Grrr! My bad...thanks for fixing this!!

services/vitalsService/getClosestValue.js
```diff
@@ -0,0 +1,110 @@
...
+        _.set(options, 'sort.vitalsTakenDateTime', 1);
+        service.find('Vital', {
+            query,
+            options,
+        }, cb);
+    }
+    function afterDateTimeProcess(res, cb) {
+        if (_.isEmpty(res) && _.isEmpty(LISTOFVITALS)) return cb(serviceUtils.bail(['-2', 'No records found']));
+        cb(null, res[0]);
+    }
```
mfuroyama: Hmm, `beforeDateTimeProcess` and `afterDateTimeProcess` look suspiciously similar! Like, exactly similar. Consider DRYing this out of your code.

services/vitalsService/getClosestValue.js
```diff
@@ -0,0 +1,110 @@
...
+    // Search flag set to 1, look for vitals with vitalsTakenDateTime before dateTime
+    function beforeDateTimeQuery(cb) {
+        _.set(query, 'vitalsTakenDateTime.$lte', searchDateTime);
+        _.set(options, 'sort.vitalsTakenDateTime', -1);
+        service.find('Vital', {
+            query,
+            options,
+        }, cb);
+    }
```
mfuroyama: Hmm...`beforeDateTimeQuery` looks suspiciously similar to `eitherDateTimeBeforeQuery`...like, exactly similar...like verbatim! Consider DRYing this out of your code.

services/vitalsService/getClosestValue.js
```diff
@@ -0,0 +1,110 @@
...
+    function eitherDateTimeAfterQuery(res, cb) {
+        if (!_.isEmpty(res)) LISTOFVITALS.push(res[0]);
+        _.unset(query, 'vitalsTakenDateTime.$lte');
+        _.set(query, 'vitalsTakenDateTime.$gte', searchDateTime);
+        _.set(options, 'sort.vitalsTakenDateTime', 1);
+        service.find('Vital', {
+            query,
+            options,
+        }, cb);
+    }
```
mfuroyama: So, this looks extremely similar to `afterDateTimeQuery` but I see what you're doing here: You've made this **special** function to handle the case where things were added to the `query` object by the `eitherDateTimeBeforeQuery` function.

Instead of creating a specialized function for this case, consider **cloning** the `query` object within each function (i.e. `beforeDateTimeQuery` and `afterDateTimeQuery`) This way, you don't need to worry about when a function was called or in what order.

There are several ways to clone-and-set objects in JS, but my technique of choices uses the lodash `_.extend` method:
```Javascript
const queryObj = _.extend({}, query, { vitalsTakenDateTime: { $gte: searchDateTime }});
```

emulation/models/gmv-closest-reading.js
creindle: Nooooooo!

emulation/models/gmv-closest-reading.js
creindle: Noooooo, why did I miss all of this :(

emulation/models/gmv-closest-reading.js
creindle: I need to add more to the model validity check that it has the service name too...

services/vitalsService/getClosestValue.js
creindle: Good thing there was a second pair of eyes on this!

PR #255, opened 2018-03-03
Vam 542 vitals write

mfuroyama:


services/tests/writeVitals-spec.js
```diff
@@ -0,0 +1,131 @@
...
+            if (err) fail(err);
+            expect(res).toBeDefined();
+            expect(res).toEqual(jasmine.any(Object));
+            expect(Object.keys(res).length).toEqual(15);
+            done();
+        });
+    });
+
+    // removeVital tests
+    xit('removeVital should correctly populate the vital properties isRemoved & removalDetails', (done) => {
```
creindle: The rest of the code down was from Adams approved PR for write Allergy spec.

services/vitalsService/createVital.js
```diff
@@ -0,0 +1,198 @@
...
+    /* {
+            vitalsTakenDateTime: Mon Feb 12 2018 21:47:14 GMT+0000 (UTC),
+            vitalType: '120_51-1',
+            hospitalLocation: '44-6',
+            value: '120/80',
+        } */
+    // I know the new vital obj will have an id based on how many vitals a particular patient has, but I must be tired because I can't figure out how to get this number and attach it to the id of the new vital object.
+    newVitalObj.id = _.get(dataObj, 'vitalType', '');
+    newVitalObj.type = 'Vital';
+    newVitalObj.vitalsTakenDateTime = { value: moment(dataObj.vitalsTakenDateTime).toISOString(), type: 'xsd:dateTime' };
```
creindle: This is giving the moment depreciation errors, but what is the best way to format this? As I saw the result had an ISO time without milliseconds.

services/vitalsService/createVital.js
```diff
@@ -0,0 +1,198 @@
...
+    // process the GmrvVitalType query
+    const gmrvVitalTypeProcess = (vitalTypeData, cb) => {
+        const vuidNum = _.get(vitalTypeData[0], 'vuid', '');
+        const vitalType = {
+            id: dataObj.vitalType,
+            label: _.get(vitalTypeData[0], 'label', ''),
+            sameAs: `vuid:${vuidNum}`,
+        };
+        newVitalObj.vitalType = vitalType;
+        newVitalObj.vitalsEnteredDateTime = { value: moment().toISOString(), type: 'xsd:dateTime' };
```
creindle: This is the same issue with the comment above at `vitalsTakenDateTime`.

services/vitalsService/createVital.js
```diff
@@ -0,0 +1,198 @@
...
+        service.find('HospitalLocation', {
+            query: { localStationNumber: context.stationId, id: hospitalLocationURN },
+            projection: serviceUtils.createProjection(['label']),
+        }, cb);
+    }
+    // process the HospitalLocation query because hospital label is not in context
```
creindle: Are we going to add the hospital label to the `sessionContext`? If not, I'll leave this here...

services/vitalsService/createVital.js
```diff
@@ -0,0 +1,198 @@
...
+            CIRCUMFERENCE_GIRTH: {
+                units: 'in',
+                metricUnits: 'cm',
+                metricValue: toString(2.54 * Number(dataObj.value), 2),
+            },
+            'PULSE OXIMETRY': {
+                units: '%',
+            },
+        };
+        newVitalObj.enteredBy = { id: `200-${serviceUtils.getIdFromUrn(context.userId)}`, label: context.userLabel };
```
creindle: I know I need to have this formatted so that the service function do not deal with IENs at all.

services/vitalsService/createVital.js
```diff
@@ -0,0 +1,198 @@
...
+                    low: '100/60',
+                },
+            }
+        */
+        cb(null);
+    }
+    // write the new Vital object to MongoDB
+    const writeVital = (cb) => {
+        console.log(newVitalObj);
+        service.create('Vital', newVitalObj, cb);
```
creindle: I'm so sad this doesn't work... it gives me this error: ```Failed: Vital validation failed: vitalsEnteredDateTime: Cast to Date failed for value "{ value: '2018-03-04T05:19:48.675Z', type: 'xsd:dateTime' }" at path "vitalsEnteredDateTime", vitalsTakenDateTime: Cast to Date failed for value "{ value: '2018-03-04T02:43:12.000Z', type: 'xsd:dateTime' }" at path "vitalsTakenDateTime", localStationNumber: Path `localStationNumber` is required.```

services/vitalsService/deleteVital.js
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env node
```
creindle: `deleteVital`, `removeVital`, and `unremoveVital` were all grabbed from Adams approved WRITE ALLERGY PR, since they should be doing the same thing, but deleting from different collections.

creindle:
A lot of the code was

services/vitalsService/createVital.js
```diff
@@ -0,0 +1,198 @@
...
+    const hospitalLocationURN = serviceUtils.createURNFromParams('44', _.split(dataObj.hospitalLocation, '-')[1], context.stationId);
+    let newVitalObj = {};
+    /* {
+            vitalsTakenDateTime: Mon Feb 12 2018 21:47:14 GMT+0000 (UTC),
+            vitalType: '120_51-1',
+            hospitalLocation: '44-6',
+            value: '120/80',
+        } */
+    // I know the new vital obj will have an id based on how many vitals a particular patient has, but I must be tired because I can't figure out how to get this number and attach it to the id of the new vital object.
+    newVitalObj.id = _.get(dataObj, 'vitalType', '');
```
creindle: I'm pretty 99% sure this is why I'm not able to enter this newly created vital object into the collection.

services/vitalsService/createVital.js
```diff
@@ -0,0 +1,198 @@
+'use strict';
+
+const _ = require('lodash');
+const moment = require('moment');
+const serviceUtils = require('../serviceUtils');
+
+/**
+ * @summary Writeback vitals clinical service functionality
+ * @param  {String}  startDate  Start date/time for vital retrieval in ISO format
+ * @param  {String}  endDate  Stop date/time for vital retrieval in ISO format.
```
mfuroyama: Let's remove these two strings here and add the definition/documentation for the `dataObj` object. The documentation should look something like this:
```Javascript
 @param  {Object} dataObj  The vital object to save
 @ param {String} dataObj.vitalsTakenDateTime  Date-time string of the vital measurement in ISO format
... // and so on
```

services/vitalsService/createVital.js
```diff
@@ -0,0 +1,198 @@
...
+    const service = this;
+    const REQUIRED_INPUT_PROPERTIES = [
+        'vitalsTakenDateTime',
+        'vitalType',
+        'hospitalLocation',
+        'value',
+    ];
+    const validInput = REQUIRED_INPUT_PROPERTIES.every((property) => {
+        if (!Object.keys(dataObj).includes(property)) {
+            requiredProp = property;
```
mfuroyama: `requiredProp` needs to be defined in the lexical scope, most likely with a `let` statement

services/vitalsService/createVital.js
```diff
@@ -0,0 +1,198 @@
...
+            return false;
+        }
+        return true;
+    });
+    if (!dataObj) callback(null, { failure: `Failed to create Vital (${requiredProp} required)`});
```
mfuroyama: I think you meant this to be `validInput`, not `dataObj`

services/vitalsService/createVital.js
mfuroyama: This should probably be:
```Javascript
newVitalObj.vitalsTakenDateTime = dataObj.vitalsTakenDateTime;
// or
newVitalObj.vitalsTakenDateTime = new Date(dataObj.vitalsTakenDateTime);
```
assuming the `vitalsTakenDateTime` is an ISO date string. Remember, the Vital objects `vitalsTakenDateTime` fields were just Date()). When in doubt, refer back to the Vitals schema in `dataAccess/mongoDB/schemas/Vital.json`

services/vitalsService/createVital.js
mfuroyama: No, no...when you 'create' an object in a collection using the `service.create` method, the `id` field is automatically filled in for you. No need to set this. You **will** need to set the vitalType id tho...

services/vitalsService/createVital.js
```diff
@@ -0,0 +1,198 @@
...
+            vitalsTakenDateTime: Mon Feb 12 2018 21:47:14 GMT+0000 (UTC),
+            vitalType: '120_51-1',
+            hospitalLocation: '44-6',
+            value: '120/80',
+        } */
+    // I know the new vital obj will have an id based on how many vitals a particular patient has, but I must be tired because I can't figure out how to get this number and attach it to the id of the new vital object.
+    newVitalObj.id = _.get(dataObj, 'vitalType', '');
+    newVitalObj.type = 'Vital';
+    newVitalObj.vitalsTakenDateTime = { value: moment(dataObj.vitalsTakenDateTime).toISOString(), type: 'xsd:dateTime' };
+    moment(dataObj.vitalsTakenDateTime).toISOString()
```
mfuroyama: Whatcha doin' here? This does nothing... :)

creindle: OH NO. So I was alternating back and forth between the different formats of date and so that was some left over code. Sorry to comment, just saw that and had to intervene.

services/vitalsService/createVital.js
```diff
@@ -0,0 +1,198 @@
...
+    newVitalObj.patient = { id: `2-${serviceUtils.getIdFromUrn(context.patientId)}`, label: context.patientLabel };
+    const vitalTypeId = dataObj.vitalType.replace('-', ':');
```
mfuroyama: LOL! Nice try here, but Vital Type **should** already be in URN form here. Remember that we only want to deal with URNs in the service functions.

Just a look-ahead tip: You can create national-type URNs with the `createURNFromParams` utility function. Just don't pass in the optional 'station number' parameter at the end and the URN returned will be national.

services/vitalsService/createVital.js
```diff
@@ -0,0 +1,198 @@
...
+    const gmrvVitalTypeQuery = (cb) => {
+        service.find('GmrvVitalType', {
+            query: { id: `urn:national:${vitalTypeId}` },
```
mfuroyama: Just pass in `dataObj.vitalType` here, since that should already be a URN

services/vitalsService/createVital.js
```diff
@@ -0,0 +1,198 @@
...
+            projection: serviceUtils.createProjection(['id', 'abbreviation', 'vuid', 'label']),
+        }, cb);
+    };
+    // process the GmrvVitalType query
+    const gmrvVitalTypeProcess = (vitalTypeData, cb) => {
+        const vuidNum = _.get(vitalTypeData[0], 'vuid', '');
+        const vitalType = {
+            id: dataObj.vitalType,
+            label: _.get(vitalTypeData[0], 'label', ''),
+            sameAs: `vuid:${vuidNum}`,
```
mfuroyama: You can leave the `sameAs` field out. It's not needed, since it's not in the Vitals.json schema, so it's not going to be written to the database. Side note: only fields defined in the schema are written to the MongoDB. All extra fields are excluded.

services/vitalsService/createVital.js
mfuroyama: Actually this will be a part of the session context object, but as part of a `visit` object. Hold on on this one, Adam implemented his code a certain way, so we'll see. I'll get back to you on this, but you might not need to do this.

services/vitalsService/createVital.js
```diff
@@ -0,0 +1,198 @@
...
+                metricUnits: 'cm',
+                metricValue: toString(2.54 * Number(dataObj.value), 2),
+            },
+            'PULSE OXIMETRY': {
+                units: '%',
+            },
+        };
+        newVitalObj.enteredBy = { id: `200-${serviceUtils.getIdFromUrn(context.userId)}`, label: context.userLabel };
+        newVitalObj.value = dataObj.value;
+        const supplementalDataObj = vitalTypeLabel[newVitalObj.vitalType.label];
```
mfuroyama: Wait, what is this doing again?

services/vitalsService/createVital.js
mfuroyama: Yup, and same issue here. This should be:
```Javascript
newVitalObj.vitalsEnteredDateTime = moment().toISOString();
```

services/vitalsService/createVital.js
mfuroyama: If you fix the date fields specified in the comments above, and then somewhere in the code above add the following line:
```Javascript
newVitalObj.localStationNumber = context.stationId;
```

services/vitalsService/createVital.js
```diff
@@ -0,0 +1,198 @@
...
+        cb(null);
+    }
+    // write the new Vital object to MongoDB
+    const writeVital = (cb) => {
+        console.log(newVitalObj);
+        service.create('Vital', newVitalObj, cb);
+    }
+    // return the result from creating new vital
+    const processWriteVital = (res, cb) => {
+        if (!res) return cb(null, { error: `Failed to create vital ID ${res.id}.` });
```
mfuroyama: I know you were probably tired here, but there's definitely something "not good" with this line of code. I'm going to let you figure it out as an exercise, but as a hint, let's be careful about how access data fields in our objects.

services/vitalsService/deleteVital.js
mfuroyama: It was wise of you to use Adam's code because your `deleteVital`,`removeVital` and`unremoveVital` were spot on!

mfuroyama: Ooh, wait..I'd make one suggestion. Only add the `comment` field to the update object **if and only if** the `comment` parameter passed into the function is valid. Otherwise, don't add it at all.

services/tests/writeVitals-spec.js
```diff
@@ -0,0 +1,131 @@
...
+    it('createVital should create an vital and return a success message for valid inputs', (done) => {
+        vitalsService.createVital(VALID_CREATE_INPUT, VALID_TEST_CONTEXT2, (err, res) => {
+            if (err) fail(err);
+            expect(res).toBeDefined();
+            expect(res.success).toBeDefined();
+            expect(res.vitalId).toBeDefined();
+            vitalId = res.vitalId;
+            done();
+        });
+    });
```
mfuroyama: You also might want to add code that reads the vital value from the id you created and verifies valid values

services/vitalsService/createVital.js
```diff
@@ -0,0 +1,198 @@
...
+function createVital(dataObj, context, callback) {
+    const service = this;
+    const REQUIRED_INPUT_PROPERTIES = [
+        'vitalsTakenDateTime',
+        'vitalType',
+        'hospitalLocation',
+        'value',
+    ];
+    const validInput = REQUIRED_INPUT_PROPERTIES.every((property) => {
+        if (!Object.keys(dataObj).includes(property)) {
```
mfuroyama: One more minor suggestion: instead of doing an `Object.keys` in the `every` iteratee, do it once right before you create `validInput`

services/vitalsService/createVital.js
```diff
@@ -0,0 +1,198 @@
...
+                        id: '200-63',
+                        label: 'ALEXANDER,ROBERT',
+                    },
+                    value: '120/80',
+                    units: 'mm[Hg]',
+                    high: '210/110',
+                    low: '100/60',
+                },
+            }
+        */
```
mfuroyama: I'd suggest removing this comment block, because honestly, the data here is totally in the wrong format (see all the comments above)

services/tests/writeVitals-spec.js
```diff
@@ -0,0 +1,131 @@
...
+        vitalType: 'urn:national:120_51:1',
+        hospitalLocation: 'urn:vista:6100:44:6',
+        value: '120/80',
+    };
+    const BAD_CREATE_INPUT = {
+        vitalsTakenDateTime: '2018-03-04T18:06:04.493Z',
+        vitalType: 'urn:national:120_51:1',
+        hospitalLocation: 'urn:vista:6100:44:6',
+        value: '120/80',
+        foo: 'bar',
```
mfuroyama: @creindle We'll need to bolster the **POSITIVE** test cases and inputs to include: 
-  `qualifiers` and `enteredBy` fields
- Different vital types

We'll also need to create more **NEGATIVE** test cases and inputs to include
- Non-existent URNs (e.g. `vitalType: 'urn:national:120_51:10000'` or `hospitalLocation 'urn:vista:6100:44:9999999'`)

services/tests/writeVitals-spec.js
```diff
@@ -0,0 +1,160 @@
...
+        context = serviceTestUtils.beforeAllScript().context;
+    });
+
+    afterAll((done) => {
+        async.eachSeries(VALID_CREATED_VITALS_IDS, (vitalId, cb) => {
+            vitalsService.deleteVital(vitalId, VALID_TEST_CONTEXT2, cb);
+        }, done);
+    });
+
+    VALID_CREATE_INPUTS.forEach((vital) => {
```
mfuroyama: Hey Christie! Good job expanding the tests out!

Could you do me a solid though and not only check that the results are defined, but that the results **values** are actually coming back as expected? Thanks!

PR #256, opened 2018-03-03
Vam 541 implement write for problem service

sidjtd:
Specs are not very fleshed out but PR-ing so all can get eyes on it!


services/problemService/createProblem.js
```diff
@@ -0,0 +1,163 @@
...
+    'snomedCTConceptCode',
+    'snomedCTDesignationCode',
+    'uniqueTermRequested',
+    'uniqueTermRequestComment',
+]);
+
+/**
+ * @summary Creates new Problem and assigns it to current patient.
+ * @param {Object} arg Object with all Problem properties to be created.
+ * @param {String} arg.diagnosis Label of Problem
```
mfuroyama: Let's add documentation here for all the allowable parameters (noting *optional* status  via =, e.g. `{String=}`) here

services/problemService/createProblem.js
```diff
@@ -0,0 +1,163 @@
...
+ * @memberof ProblemService
+ * @instance
+ */
+
+
+function createProblem(arg, context, callback) {
+    let obj = {};
+    const service = this;
+    const now = moment().toISOString();
+    const enteredByObj = { id: context.userId, label: context.userLabel };
```
mfuroyama: Minor readability suggestion: you might want to move this declaration of the `enteredByObj` closer to line 100, where it's actually used.

services/problemService/createProblem.js
mfuroyama: Also, correct me if I'm wrong, but shouldn't `diagnosis` be a URN and not a label? The emulator model documentations says the diagnosis value gets passed in with an IEN value, which gets converted to a URN. We should probably reflect this in the comments.

services/problemService/createProblem.js
```diff
@@ -0,0 +1,163 @@
...
+    let missingProp = '';
+    const isValid = REQUIRED_INPUT_PROPS.every((property) => {
+        missingProp = property;
+        return (_.has(arg, property));
+    });
+    if (!isValid) return callback(null, { failure: `Failed to create Problem (${missingProp} required)` });
+
+    /* Check that diagnosis input is valid */
+    const queryIcdDiagnosis = (cb) => {
+        const queryArg = _.get(arg, 'diagnosis.id', '');
```
mfuroyama: I think here, `diagnosis.id` should just be `diagnosis`, since the JSDocs up top say `diagnosis` is a `String` type field.

services/problemService/createProblem.js
```diff
@@ -0,0 +1,163 @@
...
+This implementation currently does not require PRIORITY as a required property in the incoming argument object since neither
+the code nor any documentation show it to be a required field. This conflicts with CPRS's performance where a PRIORITY value
+(of either ACUTE or CHRONIC) must be selected in order to create a new Problem.
+
+In order to make PRIORITY a required property, include the string 'priority' in the REQUIRED_INPUT_PROPS array below
+and ensure that the value selected is always either ACUTE OR CHRONIC. Remove 'priority' from the PROPS array.
+*/
+
+
+const REQUIRED_INPUT_PROPS = ['diagnosis'];
```
mfuroyama: There should be **way** more required input properties here. If I'm not mistaken, all of the fields in the `listMemberParameters` array that are flagged as `required` should probably be in here:

https://github.com/vistadataproject/VDM/blob/master/prototypes/problems/rpcEmulatorProblemModel.js#L59

services/problemService/createProblem.js
```diff
@@ -0,0 +1,163 @@
...
+
+const PROPS = REQUIRED_INPUT_PROPS.concat([
+    'icdd',
```
mfuroyama: According to the MVDM Problem model, fields like `icdd`, `snomedCTConceptValue` etc, should be determined/calculated in-code and not passed in via the input object argument.

mfuroyama:
In general for the `createProblem` service function, a better overall approach might be:

1. Check that the input argument object has all the required fields per the VDM `rpcEmulatorProblemModel`
2. Start a temporary `writeObject` that has all the basic information from the input transferred over.
3. Query all the necessary fields that need querying
4. From the query results, process the 'calculated' fields in the `writeObject`
5. Write the object to the 'Problem' collection

mfuroyama:
You know what? I'm going to spend a few minutes re-working the create and update code..

services/problemService/createProblem.js
```diff
@@ -0,0 +1,163 @@
...
+        if (res[0]) obj.providerNarrative = res[0];
+        cb(null, false);
+    };
+
+    /* Write new Problem object to MongoDB */
+    const writeProblem = (res, cb) => { service.create('Problem', obj, cb); };
+
+    /* Return success or failture message */
+    const finalizeWriteProblem = (res, cb) => {
+        if (!res) cb(null, { error: `Failed to create Problem ID: ${res.id}.` });
```
mfuroyama: This is no good, right? if `res` is null, printing out `res.id` will crash here

mfuroyama:
Hey Sid!

I'm really sorry, I wasn't more clear about the task I asked you to do with the test porting. I've updated the issue in Jira to add a little bit more detail to the task:

https://hawaiirg.atlassian.net/browse/VAM-573

Thanks!

mfuroyama:
So, remember we're creating round-trip tests from the RPC model to the service function back to the RPC model. For now, let's just focus on the 'write' RPCs (ORQQPL ADD SAVE, ORQQPL EDIT SAVE, etc.) What you *should* be doing is crafting RPC input args, just like the ones in the VDP native test cases. I don't see any of that in this pull request.

So, as a reminder, really, what you need to do is:

1) Build **native** RPC input argument arrays, based on the VDP test cases
These will look like an array of strings that look like this:
```1: 'GMPFLD(.01)="521774^R69."',
2: 'GMPFLD(.03)="0^"',
3: 'GMPFLD(.05)="^Diabetes mellitus"',
```

2) Pass those results into (currently non-existent) VICS Server RPC Emulation models. (Note that the models unitTestUtils function has a `emulate()` function that you can use which combines the `invokeRPC` and the `formatResult` function of the model you set up)

3) In the callback of `emulate()`, do the `expect` checks
See this file: https://github.com/vistadataproject/VDM/blob/master/prototypes/problems/rpcTestProblems.js for examples of native RPC inputs arguments for problems

PR #257, opened 2018-03-05
vam-599-rename-patientMeta-listPatients

pcchanghawaii:
rename patientMetaService to patientService, listPatientNames to listPatients and update combat vet


services/tests/getPatientVaa-spec.js
```diff
@@ -14,7 +14,7 @@ describe('lookup patient VAA spec', () => {
     it('Should return empty for the VAA patient', (done) => {
         service.getPatientVaa('urn:vista:999:2:19', context, (err, res) => {
             if (err) {
-                fail(err.toString());
+                serviceTestUtils.handleError(err, done);
```
mfuroyama: Hey Pei!

This is awesome! Could I ask one huge favor? For these calls to `serviceTestUtils.handleError`, could we change these to:
```Javascript
return serviceTestUtils.handleError(err, done);
```

At this point, we're erroring out anyway, so it doesn't make sense to continue on with the remaining Jasmine tests. The added benefit is that we don't get extra errors from all the failed references off of the result object. Thanks!!

PR #268, opened 2018-03-10
VAM-501: GMV V/M ALLDATA

creindle:
[X] All specs pass

To Do: 
[X] Calculate BMI and put into result string
[X] Ensure qualifier labels are in result string
[X] Ensure order of values in result string match exactly with JIRA details
[] Ensure string matches EXACTLY with what is written in capture files
[] Match hour, minutes, second portion of result string - Minute and second is different in DB than spec file
[] Figure out all missing values (Putting 1st spec as example of what is missing):
Current string output: 
```javascript
'02-01-18^07:07:09^98.6-^70-^13-^96-^RA^Sit^115/75-^165- ^75^^0^^^4- ^ ^32- ^81.28^0-^^CLINICX^ALEXANDER,ROBERT^Vitals'
```
Needs to match with (Identify missing/error values with parenthesis): 
```javascript
'(02-21-18)^(01:02:03)^98.6- ^70- ^13- ^96- (-)-^115/75- (RA)^165- ^(74.84)^(23.72^^)^32- ^81.28^(4.0*^2.9*^^)^0^^^CLINICX^ALEXANDER,ROBERT^Vitals'
```
What I current think missing/error values are (please correct me if I am wrong):
- 02-21-18 - I do not see entry in DB with this specific time stamp
- 01:02:03 - Entry in DB has different time from this
- RA - qualifier label (There are 2 qualifiers in DB, so I added both)
- 74.84 - weight value not rounded
- 23.72 - This might be Height actually, but I did not have an entry for height within specific timeframe, only had 1 entry for 2017-02-01
- 4.0*^2.9* - I'm still not sure about this...


emulation/models/gmv-vm-alldata.js
```diff
@@ -0,0 +1,58 @@
...
+        returnArr[2] = res.division;
+        returnArr[3] = `${res.dateRange.startDate}${res.dateRange.startDateMinutes} - ${res.dateRange.endDate}${res.dateRange.endDateMinutes}`;
+        if (_.isEmpty(res.vitals)) returnArr.push('NO DATA');
+        else {
+            res.vitals.forEach((vital) => {
+                returnArr.push(`${vital.join('^')}^Vitals`);
+            });
+        }
+        returnArr.push('');
+        console.log(returnArr);
```
mfuroyama: Eventually, let's pull this console.log out of the code

emulation/models/tests/gmv-vm-alldata-extra-spec.js
```diff
@@ -45,13 +44,13 @@ xdescribe('GMV V/M ALLDATA', () => {
...
         patientId: patientMainVitalsURN,
-        stationNumber: 9998,
+        stationId: 9998,
         userId: 'urn:vista:9998:200:63',
     };
 
     const noVitalsContext = {
         patientId: patientNoVitalsURN,
-        stationNumber: 9998,
+        stationId: 9998,
```
mfuroyama: Argh! Thanks for fixing this...

mfuroyama:
Actually, you don't need to create a new service function! You can use the `listVitals` call for the given patient and the given date range, no grouping.  As a reminder, here are the steps I left in the Jira issue for how this RPC should be implemented:

1. Use the `listVitals` for a patient using the date range, no grouping
2. Iterate through the vitals, collect all vital qualifiers and query the metadata to get abbreviations
3. query patient info (for result header info)
4. group vitals by date (the easiest way would be to use the lodash `_.groupBy` function)
5. build the output strings

services/vitalsService/listRangedVitals.js
```diff
@@ -0,0 +1,103 @@
...
+    const service = this;
+    const patientVitalsInfo = {
+        patientInfo: [],
+        unitAndRoom: ['Unit: ', 'Room: '],
+        division: 'Division: ',
+        dateRange: {
+            startDate: listOptions.startDate ? moment(listOptions.startDate).format('MMM D,YYYY').toUpperCase() : '',
+            startDateMinutes: (listOptions.startDate && moment(listOptions.startDate).format('HH:mm')) !== '00:00' ? `@${moment(listOptions.startDate).format('HH:mm:ss')}` : '',
+            endDate: listOptions.endDate ? moment(listOptions.endDate).format('MMM D,YYYY').toUpperCase() : '',
+            endDateMinutes: (listOptions.endDate && moment(listOptions.endDate).format('HH:mm')) !== '00:00' ? `@${moment(listOptions.endDate).format('HH:mm:ss')}` : '',
```
mfuroyama: Let's just collect the raw information in the service function, and do the formatting in the RPC `formatResult` function

services/vitalsService/listRangedVitals.js
```diff
@@ -0,0 +1,103 @@
...
+                data.push(hospitalLocation);
+                data.push(enteredBy);
+                return data;
+            });
+            patientVitalsInfo.vitals = vitalsData;
+        }
+        service.find('Patient', {
+            query: { id: context.patientId, localStationNumber: context.stationId },
+        }, cb);
+    }
```
mfuroyama: So, here in the service function, let's just collect all the vitals data.  Let's do all the formatting in the `formatResult` function in the RPC model

services/vitalsService/listRangedVitals.js
```diff
@@ -0,0 +1,103 @@
...
+    function vitalQualifiersQuery(res, cb) {
+        if (!_.isEmpty(res)) {
+            patientVitalsInfo.patientInfo[0] = res[0].name;
+            patientVitalsInfo.patientInfo[1] = res[0].socialSecurityNumber.slice(-4);
+            patientVitalsInfo.patientInfo[2] = moment(res[0].dateOfBirth).format('MMM D,YYYY').toUpperCase();
+            patientVitalsInfo.patientInfo[3] = `${moment().diff(res[0].dateOfBirth, 'years', false)} (Yrs)`;
+            patientVitalsInfo.patientInfo[4] = res[0].sex;
+        }
```
mfuroyama: This seems a little confusing: the function is named `vitalQualifiersQuery` but we're formatting the patient info. We should probably add the vital qualifier in here somewhere

creindle: Ahhhh, that's because I needed to pull out that information...

mfuroyama: LOLWHUT?

creindle: HAHAHAHA I meant to write "I was originally going to implement that, but I implemented something else instead..."

services/vitalsService/listRangedVitals.js
```diff
@@ -0,0 +1,86 @@
...
+            patientVitalsInfo.qualifier = _.groupBy(res, 'qualifier');
+        }
+        cb(null, patientVitalsInfo);
+    }
+    const allFunctions = [
+        listVitalsQuery,
+        patientInfoQuery,
+        vitalQualifiersQuery,
+        vitalQualifierProcess,
+    ];
```
mfuroyama: Hmm, consider re-arranging the functionality in the sub-functions a little, so that the logical grouping of the functionality is cohesive. This would add to the readability. The way it is right now is okay - it works - but some of the code doesn't match the name of the function the code lives in. Don't be afraid to break the **query** and **process** functions up into separate functions.

May I suggest the following function name list? You can kinda guess at what I'm getting at:
```Javascript
const allFunctions = [
    queryVitals,  // do the `listVitals` query here
    processVitals, // save the vitals off and collect the unique qualifiers here
    queryQualifiers, // 'GmrvVitalQualifier' query
    processQualifiers, // collect the qualifier abbreviations
    queryPatient, // pretty obvious
    processPatient, // again, pretty obvious
];
```

creindle: This is a very good suggestion! I'll fix this right now.

PR #269, opened 2018-03-11
vam-577-orwpt-ptinq

pcchanghawaii:
The implementation of this RPC is partial as there are quite many lookup fields whose schema are not there in  Mongo yet and this RPC is a complicated report that requires patient data from all over.  The structure is complete but I need to keep padding the report content and convert/verify formats as appropriate (hopefully can be done tonight). Remarks noted in code.  Failed tests denote those fields that require attentions for scoping's purpose.


pcchanghawaii:
I will hard-coded those missing fields and get back to you in one hour or so. Thanks,!

mfuroyama:
Thanks Pei!

PR #274, opened 2018-03-22
VAM-593: Vitals: Deleting Height causes BMI to be returned as zero

creindle:


emulation/models/orqqvi-vitals.js
```diff
@@ -73,6 +73,7 @@ const RPC_EMULATOR_MODEL = [{
             const output = [];
             const vitalType = _.get(vital, 'vitalType.label', '');
             const vitalValue = _.get(vital, 'value', '');
+            if (!vitalValue) return undefined;
```
mfuroyama: To guard against the case where `vital.value` might be a numeric 0, consider the following:
1. Remove the default value from the call to `_.get(vital, 'value');`
2. Use an `_.isNil` check on `vitalValue`

services/vitalsService/lookupLatest.js
```diff
@@ -52,7 +52,7 @@ function lookupLatest(startDate, endDate, context, callback) {
 
         if (vitalTakenTime) {
```
mfuroyama: If I'm not mistaken, we shouldn't even be adding the BMI value to the list if the calculation is invalid. If this is true, let's add code here to check for that case before generating a BMI entry object.

PR #276, opened 2018-03-22
Vam 518 591

kekuso:
- updated allergyService and model tests for marking an Allergy as "Entered in Error" without a comment
- forced all AllergyAssessment IENs to match patientIEN values
- cleaned up tests and docs


emulation/models/tests/orwdal32-save-allergy-spec.js
```diff
@@ -127,9 +129,53 @@ describe('ORWDAL32 SAVE ALLERGY asynchronous specs', () => {
         const res = unitTestUtils.beforeAllScript(rpcEmulatorModel);
         context = res.context;
         rpc = res;
+        // Clear the Allergy and AllergyAssessment collections for testing purposes
+        MongoDBDataAccess.clear('Allergy', (err, result) => {
+            if (err) console.log(err);
+        });
+        MongoDBDataAccess.clear('AllergyAssessment', (err, result) => {
+            if (err) console.log(err);
+        });
```
mfuroyama: So, don't forget here that these are asynchronous calls, so this might not work out like you expect. The way it is now, depending on the state of the MongoDB, `beforeAll` might not actually finish the `clear` operation. This is not unlike the other issue we had with running asynchronous setup stuff before.

What you'll probably want to do here is add a `done` callback to the `beforeAll` definition, then chain the `clear` calls via the callback. In the last callback, call the `done`. So:

```Javascript
beforeAll((done) => {
    ...
    MongoDBDataAccess.clear('Allergy', (err, result) => {
        if (err) console.log(err);
        MongoDBDataAccess.clear('AllergyAssessment', (err, result) => {
            if (err) console.log(err);
            done();
        });
    });
});

File: emulation/models/tests/orwdal32-save-allergy-spec.js [hunk 2]
```diff
@@ -127,9 +129,53 @@ describe('ORWDAL32 SAVE ALLERGY asynchronous specs', () => {
...
+
+    afterAll(() => {
+        // Cleanup the Allergy and AllergyAssessment collections
+        MongoDBDataAccess.clear('Allergy', (err, res) => {
+            if (err) console.log(err);
+        });
+        MongoDBDataAccess.clear('AllergyAssessment', (err, res) => {
+            if (err) console.log(err);
+        });
+    });
```
mfuroyama: Same thing here as the above comment. You'll probably want to redefine `afterAll` with a `done` callback, then chain the `clear` calls to take the asynchronous execution model into account. 

```Javascript
afterAll((done) => {
    ...
    MongoDBDataAccess.clear('Allergy', (err, result) => {
        if (err) console.log(err);
        MongoDBDataAccess.clear('AllergyAssessment', (err, result) => {
            if (err) console.log(err);
            done();
        });
    });
});
```

services/tests/allergyService-spec.js
```diff
@@ -28,15 +28,15 @@ describe('Allergy Service test specs', () => {
...
+        userId: 'urn:vista:6100:200:64',
         userLabel: 'ALEXANDER,ROBERT',
-        patientId: 'urn:vista:999:2:25',
+        patientId: 'urn:vista:6100:2:25',
         patientLabel: 'CARTER,DAVID',
-        divisionId: 'urn:vista:999:4:2957',
+        divisionId: 'urn:vista:6100:4:2957',
         divisionLabel: 'VISTA HEALTH CARE',
-        stationId: '999',
+        stationId: '6100',
```
mfuroyama: Ahh, so for all these, we'll need to change these back from `6100` to `999`. In the new VM (and here on out), the `6100` station ID no longer exists. Everything is `999` now. Sorry about that! Thanks!

emulation/models/tests/orwdal32-save-allergy-spec.js
kekuso: doh, forgot about the async jasmine D!

PR #280, opened 2018-03-22
VAM-605 Categorize Patient/Meta RPCs accordingly

sidjtd:
Apply appropriate labels to models and then renamed model files to consistently have appended number as part of previous word (like 'model1-rpc', as oppose to 'model-1-rpc')


sidjtd:
3 Spec Files commented out until some Clinical Collections (Problems & Vitals) are back in mongo

PR #287, opened 2018-03-28
Vam 619 create script to selectively restore mongo db collections

kekuso:
Pls test! Added optional collection params to restore only specified collections!
`npm run restore-mongo -- -h [hostname(optional)] -p [port(optional)] -c ['collectionName(s)'(optional)]`

Example:
`npm run restore-mongo -- -h 10.2.2.222 -p 27017 -c 'Drug Title Patient'`


mfuroyama:
I realize that this is after-the-fact, but how hard would it be to change the way we specify the target hostname/IP address from a hard-coded `10.2.2.222` to a scripted option (with default set to 10.2.2.222)? Let's talk about this tomorrow

scripts/restoreMongoDB.sh
```diff
@@ -3,27 +3,52 @@
...
-mongo 10.2.2.222:27017/vam --eval "db.dropDatabase()"
-
 echo "Retrieving current MongoDB export archive..."
 curl -o export.zip https://demoservices.vistadataproject.info/files/data/VICSServer/mongodb/export/current/VICSServer-MongoDB-export.zip
 unzip export.zip
 
-echo "Loading data into MongoDB (this might take a while)..."
-mongorestore ./export --host 10.2.2.222:27017
+# set default host/port
+HOST=10.2.2.222
```
mfuroyama: Ahh...sorry, man...minor thing. Now that we're "Docker-fied", can we change the default to `localhost` here, since most of the time, this will be the host we target

kekuso: sure np!

PR #290, opened 2018-03-29
Include rpcArg and context field comparison to relevant models

sidjtd:
If any of you guys nice your CPRS domain RPC's missing that uses patient ID (DFN) LMK!!!!


emulation/models/orqqpl-check-dup.js
```diff
@@ -25,26 +26,34 @@ const RPC_EMULATOR_MODEL = [
...
-            const acceptableFilters = ['R', 'A', 'I'];
-            const activeFilter = (_.includes(acceptableFilters, rpcArgs[0])) ? rpcArgs[0] : '';
+            // rpcArgs[0] is 'PATIENT ID' and not to be passed into service function to avoid other users searching other user lists
+            // However, patientId will be checked to ensure incoming rpcId matches context patientId
+            const patientId = rpcArgs[0];
+
+            const idMatch = securityService.checkIfNativeArgMatchesSession(patientId, session);
+
+            if (_.has(idMatch, 'error')) {
+                return callback(null, idMatch);
```
mfuroyama: Let's signal via the callback that some kind of error occurred here, by calling the `callback` with an Error object as the first parameter. The error object can be based on the `idMatch` field, whatever that may be.

services/securityService/authenticateRPCIdArg.js
```diff
@@ -0,0 +1,37 @@
...
+ * @memberof securityService
+ * @instance
+ */
+
+function checkIfNativeArgMatchesSession(incomingPatientId, context) {
+    const nativeArg = parseInt(incomingPatientId, 10);
+    let patientId = _.get(context, 'patientId', false);
+
+    if (!patientId) {
+        return { error: 'The context object is missing a patient Id' };
```
mfuroyama: Here, instead of returning a JSON object with an `error` field, let's return an Error object instead.

This would map better to what we eventually want to do, which is have an enumerated Error service which returns Errors based on name/code selection.

sidjtd: Got it!

emulation/models/orqqvi-vitals.js
```diff
@@ -44,6 +45,15 @@ const RPC_EMULATOR_MODEL = [{
...
     invokeRPC(rpcArgs, session, callback) {
+        // rpcArgs[0] is 'PATIENT ID' and not to be passed into service function to avoid other users searching other user lists
+        // However, patientId will be checked to ensure incoming rpcId matches context patientId
+        const patientId = rpcArgs[0];
+
+        const idMatch = securityService.checkIfNativeArgMatchesSession(patientId, session);
+
+        if (_.has(idMatch, 'error')) {
+            return callback(null, idMatch);
+        }
```
mfuroyama: Same here: let's check for a JS error object instead of an error field

PR #291, opened 2018-03-30
vam-594-sensitive-patient

pcchanghawaii:
MVDM-43-0.json should only be loaded when you need to enable the MasParameter setting. File 43 should be dumped if we we don't need the MasParameters setting.


pcchanghawaii:
merge the newer pull request (same files replicated as a new branch)

PR #293, opened 2018-04-04
Vam 636 637 638 639 enumerated errors

kekuso:
created Enumerated Errors module, updated Allergy and Vitals Service Errors


services/serviceErrorService.js
```diff
@@ -0,0 +1,32 @@
+'use strict';
```
mfuroyama: Let's rename this to `serviceError` which makes more sense and is less confusing

services/baseService.js
```diff
@@ -270,6 +271,43 @@ class BaseService extends EventEmitter {
...
+        if (typeof input === 'number') {
+            // create an error based on the service-specific id field
+            index = _.findIndex(service.errorMap, ['id', input]);
+        }
+        if (typeof input === 'string') {
+            // create an error based on the service-specific name field
+            index = _.findIndex(service.errorMap, ['name', input]);
+        }
+
+        if (index < 0) return { error: `Service Error ${input} does not exist.` };
```
mfuroyama: Let's return a generic JS Error object here

services/serviceErrorService.js
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Service for extension on JS Error Object.
+ *
+ * @class ServiceErrorService
+ */
+class ServiceErrorService extends Error {
```
mfuroyama: Let's change the name of this class to `ServiceError`, and rename the file accordingly

services/baseService.js
```diff
@@ -270,6 +271,43 @@ class BaseService extends EventEmitter {
...
+
+        if (optionalArg) return new ServiceError(service.errorMap[index], this.domainName, optionalArg);
```
mfuroyama: If I'm not mistaken, we probably don't need the conditional here, since the `ServiceError` class will manage the validity of the `optionalArg` parameter. Let's just:
```Javascript
return new ServiceError(service.errorMap[index], this.domainName, optionalArg);
```

services/baseService.js
```diff
@@ -1,11 +1,12 @@
 'use strict';
 
 const EventEmitter = require('events');
-const fs = require('fs');
+const fs = require('fs-extra');
 const path = require('path');
 const moment = require('moment');
 const _ = require('lodash');
 const DataAccess = require('../dataAccess');
+const ServiceError = require('./serviceErrorService');
```
mfuroyama: ...and we'll need to change this too, once we rename the `ServiceError` module file name

mfuroyama:
So, let's make it a standard that we create errors in the code using the error **name** instead of the **id**. Makes for better readability. But let's keep the option to generate errors by id in there, just in case.

services/VISTAMetaService/errors.json
```diff
@@ -0,0 +1,13 @@
+{
```
mfuroyama: Actually, let's delete this file out.

PR #297, opened 2018-04-10
Vam 582 616 implement allergy negative jasmine test cases

kekuso:
- refactor specs for Enumerated errors
- this should include the generic errors loading when a service is called


emulation/models/tests/orwdal32-save-allergy-spec.js
```diff
@@ -170,11 +171,12 @@ describe('ORWDAL32 SAVE ALLERGY asynchronous specs', () => {
             done();
         });
     });
-    it('invokeRPC: Try to mark Patient as NKA again, expect failure message', (done) => {
+    it('invokeRPC: Try to mark Patient as NKA again, expect "Redundant" message', (done) => {
         rpc.invokeRPC(NO_KNOWN_ALLERGIES, TEST_CONTEXT, (err, res) => {
             if (err) unitTestUtils.handleError(err, done);
```
mfuroyama: Ahh, so `unitTestUtils.handleError` actually calls `done` internally, so to prevent a `callback was already called` error here and everywhere, we need to do a `return` here, so:
```Javascript
if (err) return unitTestUtils.handleError(err, done);
```

services/baseService.js
```diff
@@ -337,10 +337,16 @@ class BaseService extends EventEmitter {
...
         if (!errorDefinitions.length) return;
-        const json = fs.readJsonSync(path.join(loadPath, errorDefinitions[0]));
-        _.set(classDef.prototype, 'errorMap', json.errors);
+
+        const genericDefinitions = fs.readdirSync(__dirname)
+            .filter(fileName => fileName === 'errors.json');
+        if (!genericDefinitions.length) return;
+
+        const genericJson = fs.readJsonSync(path.join(__dirname, genericDefinitions[0]));
+        const serviceJson = fs.readJsonSync(path.join(loadPath, errorDefinitions[0]));
```
mfuroyama: Actually, here, since we know the file name and path ahead of time, we don't really need to use `readdirSync` to grab `errorDefinitions` and `genericDefinitions`. We were only doing that above because we don't know the names of the service function files beforehand, so we used the `readdirSync` to dynamically collect the file names from the directory.

Instead, consider building the `errors.json` file paths from what you know, then calling `fs.readJsonSync` with the `throws` option set to false (which causes `readJsonSync` to return null if the file doesn't exist) [(docs here)](https://github.com/jprichardson/node-fs-extra/blob/master/docs/readJson-sync.md):

```Javascript
// Remove the 'errorDefinitions' and 'genericDefinitions' code above, then:
const genericJson = fs.readJsonSync(path.join(__dirname, 'errors.json'), { throws: false }) || [];
...
```

services/errors.json
```diff
@@ -1,16 +1,16 @@
 {
     "errors": [{
-        "id": 1,
+        "id": "01",
```
mfuroyama: Hmm, I think I know why you did this, but could we set these to numeric values instead (So that the generic error IDs don't clash with the service-specific errors?). Instead, let's use something crazy high, like 100001, 100002 or something? Better to be consistent and true to the object data type specification here.

services/baseService.js
```diff
@@ -335,12 +335,10 @@ class BaseService extends EventEmitter {
...
 
-        const errorDefinitions = fs.readdirSync(loadPath)
-            .filter(fileName => fileName === 'errors.json');
-
-        if (!errorDefinitions.length) return;
-        const json = fs.readJsonSync(path.join(loadPath, errorDefinitions[0]));
-        _.set(classDef.prototype, 'errorMap', json.errors);
+        const serviceJson = fs.readJsonSync(path.join(loadPath, 'errors.json'), { throws: false }) || [];
+        const genericJson = fs.readJsonSync(path.join(__dirname, 'errors.json'), { throws: false }) || [];
+        const json = genericJson.errors.concat(serviceJson.errors);
```
mfuroyama: Ahh, my bad...I thought `errors.json` only had an array of objects. It's actually an object with an `errors` array field. In that case, to ensure that this line (340) doesn't crash, best to do this:
```Javascript
const serviceJson = fs.readJsonSync(path.join(loadPath, 'errors.json'), { throws: false }) || { errors: [] };
const genericJson = fs.readJsonSync(path.join(__dirname, 'errors.json'), { throws: false }) || { errors: [] };
```

PR #299, opened 2018-04-11
Vam 617 problem negative tests

sidjtd:
Adam helped me with restoring some error handlers that I removed accidentally so hopefully I got those all right......


emulation/models/tests/orqqpl-replace-spec.js
```diff
@@ -15,28 +15,20 @@ describe('ORQQPL REPLACE asynchronous specs', () => {
...
-    });
+    afterAll((done) => { utils.unloadPatientData('Problem', 999, utils.handleDone(done)); });
 
-    const context = {
-        stationId: 999,
-    };
+    const context = { stationId: 999 };
 
-    it('Should set an existing "removed" problem to "unremoved"', (done) => {
+    xit('Update existing "removed" problem to "unremoved"', (done) => {
```
mfuroyama: Hmm, just wondering if this test spec is commented out for a reason...

services/problemService/errors.json
```diff
@@ -41,13 +41,13 @@
     },
     {
         "id": 9,
-        "name": "MISSING_GMPORIG_ENTRIES",
-        "description": "The array parameters are invalid: missing GMPORIG entries"
+        "name": "MISSING_ORARRAY_ENTRIES",
```
mfuroyama: Here, and in some of the other enumerated error names, let's try to move away from RPC/MUMPSy terms in the errors. For example, here, 'GMPORIG' and 'ORARRAY' definitely have that RPC odor about them.

services/problemService/unremoveProblem.js
```diff
@@ -12,19 +13,15 @@
...
     }, (err, res) => {
-        if (err) {
-            return callback(err);
-        }
-        if (!res) {
-            return callback(new Error(`Failed to unremove problem ID ${problemId} from the Problem collection`));
-        }
-        callback(null, { success: `Successfully unremoved Problem ID ${problemId}` });
+        if (err || _.isEmpty(res)) return callback(this.error('FAILED_UNREMOVE_PROBLEM', problemId));
+        if (res) return callback(null, { success: `Successfully unremoved Problem ID ${problemId}` });
```
mfuroyama: Here, no need for the `if (res)` conditional

mfuroyama:
@sidjtd could you help me out here? I see a bunch of changes to handle enumerated errors, and code clean-up, formatting, etc., but I don't see that many new, added negative test cases? Could you point those out to me? Thanks!

services/problemService/errors.json
sidjtd: Got it!

services/problemService/unremoveProblem.js
sidjtd: Ah, I guess the return negates the need for that 100%? I was thinking how its worrisome if callback got called twice by accident, as it seems to on a rare occasion....

mfuroyama: True you don't want to call `callback` twice, but on the other hand, we need to ensure that `callback` is called at least once.

In this specific case, there are two reasons I'd recommend removing the `if (res)` check on the last statement

1. The `if (res)` case you're checking for _should_ be covered by the guard in the previous line, so it's really unnecessary, and removing it makes the code cleaner.

2. Since `callback` **must** be called, it's best that your function have a default, fall-through case that calls `callback`. If, for some odd reason, `_.isEmpty(res)` and `(res)` are both falsy, then the execution flow will fall to the end of the function without having called `callback`.

On a side note, a personal habit I follow is to make sure that I never end a function with an open `if` statement. This helps me guarantee that there are no unexpected return cases and that I have full control over the execution flow.

sidjtd:
The majority of negative test I added covers the service functions when receiving an argument not valid/missing attributes that make it valid, an update fail, or when process in the middle of a service Function execution doesn't return a necessary piece of info to continue the process.  In other words, with a handful of minor exceptions, I only created tests for ERRORS when the rpcModel/serviceFunction already had a previous error implementation (that I replaced with the error class ver).  So... although they may not be in the model specs, they should be covered in the service Function specs? I don't think that I'm missing a lot of negative tests? But If I did though.... pls lmk!

services/problemService/unremoveProblem.js
sidjtd: I will remove the last conditional, but wouldn't callback definitely be called at least once if perhaps I used if(!_.isEmpty(res)) [has bang] to contrast with plain old _.isEmpty(res)?

Or maybe just wrap it in an "else"? I think I did that in the past quite a few times but the linter complained so much I didn't have the heart to keep it there....

mfuroyama: > but wouldn't callback definitely be called at least once if perhaps I used if(!_.isEmpty(res)) [has bang] to contrast with plain old _.isEmpty(res)?

Right, that was kinda the point of recommendation #1 above. You don't need it, so I'd suggest removing it...which, as it turns out, is also good practice anyway because of recommendation #2.

And yeah, don't wrap that last part in an `else`, just let it fall through. The linter will throw a warning if you use a meaningless `else` because, well, it's meaningless... 😄

emulation/models/orqqpl-user-prob-cats.js
```diff
@@ -33,8 +35,16 @@ const RPC_EMULATOR_MODEL = [
             example: '3',
         }],
         invokeRPC: (rpcArgs, session, callback) => {
-            const user = rpcArgs[0] ? modelUtils.createURNFromParams(200, rpcArgs[0], session.stationId) : '';
-            if (user !== session.userId) return callback(new Error('User IEN from Client does not match with session user IEN'));
+            // rpcArgs[0] is 'PATIENT ID' and not to be passed into service function to avoid other users searching other user lists
+            // However, patientId will be checked to ensure incoming rpcId matches context patientId
+            const patientId = rpcArgs[0];
+
+            const idMatch = securityService.checkIfNativeArgMatchesSession(patientId, session);
```
mfuroyama: Just a heads-up: The Security Service API has changed slightly, namely, the `checkIfNativeArgMatchesSession` function was changed to `verifyPatientId`. This was done to accommodate the additional `verifyUserId` function, which checks the NewPerson URN against user IENs passed in to an RPC

PR #300, opened 2018-04-12
VAM-642: Add provider/user IEN checks to the Security Service

creindle:
[X] Implemented userId to session.userId check 
[X] Renamed original implementation of patientId to session.patientId check
[X] Ensured all inputs with patientId and userId in the models are replaced with new functions


services/securityService/authenticateRPCIdArg.js
```diff
@@ -4,34 +4,53 @@ const _ = require('lodash');
...
+
+    if (!contextUserId) {
+        return new Error('The context object requires contextUserId field');
+    }
+
+    contextUserId = parseInt(serviceUtils.getIdFromUrn(contextUserId), 10);
+    const output = (_.isEqual(inputUserId, contextUserId) && !_.isNaN(inputUserId));
+
+    if (!output) {
+        return new Error(`User IEN from Client ${inputUserId} does not match with session user ID ${contextUserId}`);
```
mfuroyama: Let's convert these errors into enumerated errors (which means adding an `errors.json` file for the security service)

PR #302, opened 2018-04-13
Vam 645 implement initial rest meta endpoints

kekuso:
- Let me know if you want any more features/testing, but I think I got the basic idea for our REST endpoints!


mfuroyama:
I pushed a minor change that tweaks the `package.json` in your branch a little bit. It basically moves the `nodemon` package to the dev dependencies, and adds an npm script (`watch`) that we can use to activate `nodemon`

emulation/httpserver/routes/allergy/allergens.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+const LexiconService = require('../../../../services/lexiconService');
+
+const lookupAllergens = (req, res) => {
+    // Lookup Allergens by search term, e.g. "CHOCOLATE"
+    const searchTerm = req.query.search || '';
+
+    LexiconService.lookupAllergies(searchTerm, (err, result) => {
+        if (err) console.log('err: ', err);
```
mfuroyama: So for now, if the service function returns an error, let's just return a 500 error code with the Error object returned as JSON:
```Javascript
if (err) return res.status(500).json(err);
```

PR #306, opened 2018-04-17
VAM-647: Implement REST endpoints for Vital meta data

creindle:
[X] Need to add info to wiki

lookupTypeHierarchy reformatted and documented [here](https://github.com/vistadataproject/VICSServer/wiki/VICS-Server---Service-REST-API#lookup-vital-type-hierarchy)


emulation/httpserver/routes/vital/typeHierarchy.js
```diff
@@ -0,0 +1,18 @@
...
+
+    VitalMetaService.lookupVitalTypeHierarchy(vitalTypes, req.context, (err, result) => {
+        if (err) return res.status(500).json(err);
+        return res.json(result);
+    });
+};
+
+module.exports = [{
+    method: 'GET',
+    path: '/typeHierarchy',
```
mfuroyama: Hmm, what do you think about naming this endpoint `types`? Let's discuss

mfuroyama:
So, while you were implementing these new endpoints, I created a utility that allows you to test route endpoint functions **without** having to call actual service functions. The utility wraps and stubs the service functionality out, which lets you isolate and test the endpoint code only.  I also created an example against the `allergy/allergens` endpoint (`allergens-spec.js`) to demonstrate how to use the utility.

If you could, merge the latest `master` into this feature branch and implement Jasmine unit tests for your endpoint handler. Thanks!

mfuroyama:
The test looks good. I'll mark this pull request 'Approved' after the appropriate documentation is added to the wiki (https://github.com/vistadataproject/VICSServer/wiki/VICS-Server---Service-REST-API)

creindle:
The documentation is complete!

mfuroyama:
For the endpoint documentation, could you expand the documentation to include definitions for the "category" sub-objects, and the "qualifier" sub-objects? Right now, as it is, it looks like the documentation stops at the object array. Thanks!

mfuroyama:
Per our conversation, let's restructure the raw service function output and modify the documentation to reflect the new output data structure. Thanks!

PR #307, opened 2018-04-17
Vam 646 implement rest crud endpoints for allergy patient data

kekuso:


mfuroyama:
So, while you were implementing these new endpoints, I created a utility that allows you to test route endpoint functions **without** having to call actual service functions. The utility wraps and stubs the service functionality out, which lets you isolate and test the endpoint code only.  I also created an example against the `allergy/allergens` endpoint (`allergens-spec.js`) to demonstrate how to use the utility.

If you could, merge the latest `master` into this feature branch and implement Jasmine unit tests for your endpoint handler. Thanks!

emulation/httpserver/routes/allergy/lookupAllergy.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+const AllergyService = require('../../../../services/allergyService');
+
+const lookupAllergy = (req, res) => {
+    const allergyId = req.params.id || '';
+    const allergyUrn = allergyId ? `urn:vista:${req.context.stationId}:120_8:${allergyId}` : '';
```
mfuroyama: This is fine, but just a thought: we *might* want to create a utilities module for the HTTP/route stuff

PR #308, opened 2018-04-18
vam-594-sensitive-patient-2update the file to fix a copy error

pcchanghawaii:
I noticed the eslint errors but Christie was fast enough to fixed them in the master before this pull request.  So I created another new branch based on the master around 4:30pm(CA time) just to fix a copy error in my last master merge.  There is an important line missing in my last merge(lookupDgSensitiveRecord.js), which caused the errors for lookupDgSensitiveRecord10000MasEnabled-spec.js and lookupDgSensitiveRecord10000-spec.js 
Sorry about that. 

PS: lookupDgSensitiveRecord10000MasEnabled-spec.js and lookupDgSensitiveRecord10000-spec.js are not supposed to be both passed at the same time. MasEnabled should only be passed when there is a record in file 43 (MasParameters) and lookupDgSensitiveRecord10000 should only be passed when is is NONE record in file 43.

Thanks for fixing the bugs!


mfuroyama:
Awesome! Actually, I started looking at this, since I **think** the missing line was causing the regression test to fail!

Yeah, your initial push to master had ESLint errors, but Sid pushed changes over the top of yours, so it became his responsibility to fix the build at that point. Luckily, @creindle stepped in and fixed the errors.  Thanks, Pei!

PR #313, opened 2018-04-24
vam-661-stress-dataset

pcchanghawaii:
package file was modified to include one more test-mongo task.
so the script can be triggered via npm run test-mongo, which executes
node ./tests/utils/patientJsonGenerateInsert.js -n localhost
you may modify the script to increase the quantity of synthetic patient records, which is default to 10000.  Larger heap is required to run a larger quantity.  

Should I test the dataset against the following patient-related RPCs?
https://github.com/vistadataproject/VDM/tree/master/prototypes/patientVICS

ps: I've also included the static data set 10000 in one of the master test util datasets.


pcchanghawaii:
thanks for reviewing. i will merge it in the afternoon.

PR #318, opened 2018-04-26
Vam 673 create update problem endpoints

sidjtd:
updateProblem endpoint was made identically to createProblem endpoint since problemUpdate.js (service function) does not have a 2nd parameter problem-arg, unlike updateAllergies which has allergyProperties as the 2nd arg. The 2nd expected arg in problemUpdate is the context.


mfuroyama:
Code looks good, just add the API docs (separate YML) files, and we're good!

sidjtd:
Rogerrr

PR #321, opened 2018-04-27
Vam 649 prob meta res tendpoints

sidjtd:
Pretty sure index.js may be causing the "cant automatically merge" error? Or masters already has meta and patient directories? Either way, can try adjust, making PR first!


emulation/httpserver/routes/problem/problemCategories.yml
```diff
@@ -0,0 +1,34 @@
+paths:
+  /meta/problems/categories:
+    get:
+     summary: Shows 
+     description: Returns array of problem categories for user to select from
+     parameters:
+        - in: header
+          name: clinical locaiton
```
mfuroyama: If I'm not mistaken, I think this should be:
```yaml
     - in: query
       name: id  // <-- this is the name of the parameter in the code
```

emulation/httpserver/routes/problem/problemCategories.yml
```diff
@@ -0,0 +1,34 @@
...
+          description: Clinical Location ID
+          schema:
+            type: String
+            example: '32'
+     responses:
+      200:
+        description: List of vitals
```
mfuroyama: We _might_ want to change this to reflect the actual response description...

emulation/httpserver/routes/problem/problemCategories.yml
```diff
@@ -0,0 +1,34 @@
...
+        schema:
+          type: array
+          items: 
+            type: string
```
mfuroyama: If I'm not mistaken, this should be:
```yaml
schema:
    type: array
    items:
        type: object
        properties:
            label:
                type: string
                ...
```

PR #323, opened 2018-04-28
VAM-675: Implement REST-ful GET endpoints for Problem Patient Data

creindle:
[] Need to implement new endpoints


mfuroyama:
for `lookupProblem` lets change the code (and the docs) to return a single object instead of an array with a single object. Thanks

PR #327, opened 2018-04-30
Vam 646 Update Documentation to take multiple YML files across multiple files

sidjtd:
npm run doc-http is the command to run it at the moment, and the index.html is created in docs/http directory!


mfuroyama:
Wait...double check the files in this feature branch. It looks like you have some old files in there that shouldn't be in there

package.json
```diff
@@ -79,7 +79,8 @@
     "through2": "^2.0.3",
     "underscore": "^1.8.3",
     "uuid": "^3.1.0",
-    "vam-rpcparser": "git+https://github.com/vistadataproject/rpcParseFormat.git"
+    "vam-rpcparser": "git+https://github.com/vistadataproject/rpcParseFormat.git",
+    "js-yaml": "^3.11.0"
```
mfuroyama: Let's save this to the `devDependencies` instead of the regular `dependencies`

scripts/createApiHttpDocs.js
```diff
@@ -1,39 +1,78 @@
...
-const path = require('path');
-const _ = require('lodash');
 const fs = require('fs');
+const _ = require('lodash');
+const path = require('path');
+const yaml = require('js-yaml');
+
+// const emulatorRoutePath = '../emulation/httpserver/routes';
+const metaRoutePath = '../emulation/httpserver/routes/meta';
+const patientRoutePath = '../emulation/httpserver/routes/patient';
```
mfuroyama: This works fine, but instead of hard-coding the target source paths here, you might want to consider making the script slightly more robust by having it dynamically search the subdirectories in the `routes` directory. This way, the script is more resilient to change (like, for example, if we change the directory names, or if we add new route directories, etc.)

scripts/createDocsDirectory.js
```diff
@@ -3,11 +3,21 @@
...
     }
 }
 
+function removeGreatYaml(dir) {
+    const erasePath = path.join(docPath, '/z_theGreatYaml.yml');
+    console.log('erase', erasePath);
+    if (fs.existsSync(erasePath)) {
+        fs.unlinkSync(erasePath);
+    }
+}
```
mfuroyama: It looks like this is not being used, consider deleting it?

scripts/createDocsDirectory.js
```diff
@@ -3,11 +3,21 @@
 const path = require('path');
 const fs = require('fs');
 
+let docPath = '';
```
mfuroyama: If I'm not mistaken, this variable definition in the outer scope can probably go away and move back into `createDocsDirIfMissing`

Issue #331, opened 2018-05-02
id lookups do not work with urn-based id


Hokua:
they work with the last number in the urn id, example urn:vista:999:2:17 => 17.

This is true for allergyLookup and removeAllergies

mfuroyama:
Added [VAM-688](https://hawaiirg.atlassian.net/browse/VAM-688) in Jira VAM project for tracking

mfuroyama:
@kekuso ended up reworking all the endpoints to use URNs.

PR #333, opened 2018-05-03
Vam 688 standardize passing urns instead of ids in http endpoint implementations

kekuso:


mfuroyama:
@kekuso the code looks good! But, could you do me a solid? Could you go through the YAML files and update the endpoint documentation files so that it's clear where we're using URNs? Thanks!

mfuroyama:
This pull request addresses Github issue #331

PR #336, opened 2018-05-07
VAM-679: Create smoke test to ensure nodeVISTA and MongoDB are in the correct state

creindle:
Run with `npm run checkDataSync`!


package.json
```diff
@@ -38,6 +38,7 @@
     "restart-all": "npm run restart-vicsdb && npm run restart-nodevista",
     "watch": "nodemon",
     "dist": "gulp dist",
+    "checkDataSync": "./scripts/checkDataSync.sh",
```
mfuroyama: For consistency, could we change the name of this script to use kebab-case? (as in: `check-data-sync`) Thanks!!

scripts/checkDataSync.js
```diff
@@ -0,0 +1,146 @@
...
+    },
+    {
+        name: 'Allergy',
+        url: 'http://localhost:9000/fmqlEP?fmql=DESCRIBE%20120_8%20LIMIT%20100%20OFFSET%200',
+    },
+    {
+        name: 'Problem',
+        url: 'http://localhost:9000/fmqlEP?fmql=DESCRIBE%209000011%20LIMIT%20100%20OFFSET%200',
+    },
+];
```
mfuroyama: Here, consider using an object keyed by domain name instead of an array of objects. So, instead of:
```Javascript
const dataCollections = [{
    name: 'Vital',
    url: 'http://...',
}];
```

You'd do this:
```Javascript
const dataCollections = {
    Vital: 'http://....',
    Allergy: 'http://....',
};
```

I notice below that you're accessing these values using numeric array indexes. Changing `dataCollections` into a domain-keyed object and accessing the URLs by domain name would make the code a little bit more readable by expressing your intent through the code.

If you tweaked `getNodeVISTAData` to take URLs, your code below:
```Javascript
function getVitalNodeVISTAData(cb) {
    getNodeVISTAData(dataCollections, 0, cb);
}
```

Could turn into something like this:
```Javascript
function getVitalNodeVISTAData(cb) {
    getNodeVISTAData(dataCollections.Vital, cb);
}
```

In your previous example, I have no idea what `0` corresponds to, but with the second example, I immediately see that you're grabbing using something that's related to `Vitals`

package.json
creindle: Thank you for that suggestion! Makes a lot of sense.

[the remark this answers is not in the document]
creindle: That makes so much more sense! Perfect!

scripts/checkDataSync.js
creindle: That's a very good idea and does look a lot more clear!

scripts/checkDataSync.js
```diff
@@ -0,0 +1,146 @@
...
+            {
+                name: 'allergy',
+                nodeVISTAContent: formatNodeVISTAContent(res, 1),
+                mongoDBContent: formatMongoDBContent(res, 1),
+            },
+            {
+                name: 'problem',
+                nodeVISTAContent: formatNodeVISTAContent(res, 2),
+                mongoDBContent: formatMongoDBContent(res, 2),
+            },
```
mfuroyama: If I'm understanding the code correctly, you might want to consider applying the 'domain-keyed object' concept here as well.

PR #338, opened 2018-05-09
VAM-690: Add additional information/meta to RPC Engine RPC objects

creindle:


emulation/dispatchers/pureEmulation.js
```diff
@@ -150,6 +150,9 @@ class PureEmulationDispatcher {
         const handler = dispatcher.name;
         return (rpc, app, next) => {
             const name = rpc.get('name', '');
+            const model = dispatcher.emulationModelMap[name];
+            const rpcType = _.get(model, 'rpcType', '');
+            const labels = _.intersection(['META', 'PATIENT'], model.labels);
```
mfuroyama: Let's make it so that these variable assignments (and the `set` calls below) **only** happen if the `model` object we're looking for exists.

emulation/dispatchers/pureEmulation.js
```diff
@@ -150,6 +150,15 @@ class PureEmulationDispatcher {
...
         return (rpc, app, next) => {
             const name = rpc.get('name', '');
+            const model = dispatcher.emulationModelMap[name];
+            if (model) {
+                const rpcType = _.get(model, 'rpcType', '');
+                const labels = _.intersection(['META', 'PATIENT'], model.labels);
+                const domain = _.get(model, 'domain', '');
+                rpc.set({ rpcType });
+                rpc.set({ labels });
+                rpc.set({ domain });
```
mfuroyama: Let's clean the code up here a little bit by merging the 3 `set` calls into one.

creindle: Oh yeah duh!

PR #345, opened 2018-05-16
Vam 703 test creating observed allergies via create allergy res tful endpoint using vics client web app

kekuso:


services/allergyService/createAllergy.js
```diff
@@ -112,7 +112,8 @@ function createAllergy(args, context, callback) {
         allergyObj.observedOrHistorical = OBSERVED_OR_HISTORICAL_MAP[args.observedOrHistorical];
     }
     allergyObj.verified = (args.observedOrHistorical === 'HISTORICAL' ? 'YES' : 'NO');
-    allergyObj.dateOccurred = now.toISOString();
+    const dateOccurred = args.dateOccurred ? new Date(args.dateOccurred) : '';
```
mfuroyama: In this ternary, consider replacing the "empty string" secondary result with `now`. That way, you don't need the ternary in the line below

PR #350, opened 2018-05-22
VAM-706: Add the ability to create multiple vital measurements at once via VICS Server HTTP RESTful endpoint

creindle:
[X] Implemented multiple vitals (array of objects) creation while keeping original single vital creation (single object)
[X] Created new yaml file to document multiple vitals creation
[X] createVital spec file updated to check multiple vital responses are reached


mfuroyama:
I updated and revamped the HTTP API generator to 1) allow multiple paths to be defined within a single YML file 2) run from Javascript/Gulp task (no intermediate script required)

Using this update, I updated the 'addVitals' YML file to use the updated generator script.

emulation/httpserver/routes/patient/vital/createVital.js
```diff
@@ -1,12 +1,28 @@
...
 const VitalService = require('../../../../../services/vitalsService');
 
 const createVital = (req, res) => {
-    VitalService.createVital(req.body, req.context, (err, result) => {
-        if (err) return res.status(500).json(err);
-        return res.json(result);
-    });
+    const allResults = [];
+    if (_.isArray(req.body)) {
+        async.each(req.body, (eachReq, callback) => {
```
mfuroyama: Consider using `async.mapSeries` here instead of `async.each`. ([async `mapSeries` docs](https://caolan.github.io/async/docs.html#mapSeries))  By using the `mapSeries` function, you don't need to maintain the `allResults` array, since `mapSeries` will collect the results for you!

So, instead of calling `allResults.push` in the `createVital` callback, you'd just call on success:
```Javascript
callback(null, result);
```

`mapSeries` will keep store the results for you. Then, in the final callback, you can just do this:
```Javascript
}, (error, result) => {
    if (error) return res.status(500).json(error);
    return res.json(result);
})
```

In the final callback, `result` will contain an array of all the "success" results from the callbacks used in the iterator.

creindle: Awesome! I'll definitely use this!

PR #355, opened 2018-05-30
VAM-731: Add treatment factor information to the VICS Server Patient data demographics HTTP GET endpoint

creindle:
[X] HTTP passes and parses treatmentFactors in output
[X] Service (getPatientReport) updated to include treatmentFactors
[X] HTTP yml file updated
[X] Spec files for both HTTP and service functions still pass


emulation/httpserver/routes/patient/data/getPatientReport.yml
```diff
@@ -51,6 +51,31 @@ definitions:
         type: string
       periodOfService:
         type: string
+        
+  TreatmentFactors:
+    type: object
```
mfuroyama: It doesn't look like this is accurate...possibly a case of the ol' "copy-pasta"? 😉 

Let's double-check the results object schema against the current [MongoDB schema](https://github.com/vistadataproject/VICSServer/blob/master/dataAccess/mongoDB/schemas/Patient.json#L2033) for Patient treatment factors.

creindle: I believe this is accurate? This is the output of treatmentFactors 
```Javascript
"treatmentFactors": [
        {
            "radiationExposure": {
                "exposureIndicated": "YES",
                "exposureMethod": "OTHER"
            },
            "agentOrangeExposure": {
                "exposureIndicated": "NO"
            },
            "southwestAsianConditions": {}
        }
    ]
```

emulation/httpserver/routes/patient/data/getPatientReport.yml
```diff
@@ -123,6 +148,10 @@ definitions:
         type: array
         items:
           $ref: '#/definitions/MilitaryService'
+      treatmentFactors:
+        type: array
+        items:
+          $ref: '#/definitions/TreatmentFactors'
```
creindle: treatmentFactors is an array and its items are objects as so 
```Javascript
{
         "radiationExposure": {
             "exposureIndicated": "YES",
             "exposureMethod": "OTHER"
         },
         "agentOrangeExposure": {
             "exposureIndicated": "NO"
         },
         "southwestAsianConditions": {}
        }```

File: emulation/httpserver/routes/patient/data/getPatientReport.yml (hunk 1 above)
Comment (creindle) [reply emulation/httpserver/routes/patient/data/getPatientReport.yml]: I specifically matched it up to be similar to `militaryService`. Unless it would be better to make this an object with key value pairs like so:
```Javascript
{
            "radiationExposure": {
                "exposureIndicated": "YES",
                "exposureMethod": "OTHER"
            },
            "agentOrangeExposure": {
                "exposureIndicated": "NO"
            },
            "southwestAsianConditions": {}
        }```

File: emulation/httpserver/routes/patient/data/getPatientReport.js [hunk 3]
```diff
@@ -56,6 +56,15 @@ const getPatientReport = (req, res) => {
             });
         }
 
+        const treatmentFactors = result.treatmentFactors;
+        if (treatmentFactors && treatmentFactors.length) {
+            finalRes.treatmentFactors = {
+                radiationExposure: _.some(treatmentFactors[0].radiationExposure, { exposureIndicated: 'YES' }),
+                agentOrangeExposure: _.some(treatmentFactors[0].agentOrangeExposure, { exposureIndicated: 'YES' }),
+                southwestAsianConditions: _.some(treatmentFactors[0].southwestAsianConditions, { exposureIndicated: 'YES' }),
```
mfuroyama: Wait, we're still only looking at the first treatment factor object here. Let's iterate through **all** the treatment factor objects to find the appropriately set `exposureIndicated` field.

emulation/httpserver/routes/patient/data/getPatientReport.js
```diff
@@ -59,9 +59,9 @@ const getPatientReport = (req, res) => {
         const treatmentFactors = result.treatmentFactors;
         if (treatmentFactors && treatmentFactors.length) {
             finalRes.treatmentFactors = {
-                radiationExposure: _.some(treatmentFactors[0].radiationExposure, { exposureIndicated: 'YES' }),
-                agentOrangeExposure: _.some(treatmentFactors[0].agentOrangeExposure, { exposureIndicated: 'YES' }),
-                southwestAsianConditions: _.some(treatmentFactors[0].southwestAsianConditions, { exposureIndicated: 'YES' }),
+                radiationExposure: _.filter(treatmentFactors, { radiationExposure: [{ exposureIndicated: 'YES' }] }).length > 0,
+                agentOrangeExposure: _.filter(treatmentFactors, { agentOrangeExposure: [{ exposureIndicated: 'YES' }] }).length > 0,
+                southwestAsianConditions: _.filter(treatmentFactors, { southwestAsianConditions: [{ exposureIndicated: 'YES' }] }).length > 0,
```
mfuroyama: You might want to consider DRYing this out by moving the `_.filter` call out into a function, then calling the function for each field type.

emulation/httpserver/routes/patient/data/getPatientReport.js
```diff
@@ -18,6 +18,8 @@ const CONTACT_PROPS = [
     'phoneNumberCellular',
 ];
 
+const isExposureIndicated = (resTreatmentFactors, radiationType) => _.filter(resTreatmentFactors, { [radiationType]: [{ exposureIndicated: 'YES' }] }).length > 0;
```
mfuroyama: One last thing, and I think this should be good. You *might* want to consider moving the definition of this function down to **after** line 61 and refer to the `treatmentFactors` array directly in your function. This turns your function into a closure, which means you don't need to specify the `resTreatmentFactors` parameter in the function definition. Makes the code even more clean. What do you think?

creindle: That's actually a really good suggestion! I was wondering that myself, but didn't know if it was considered clean to create the function there. Makes sense though!! Thank you.

mfuroyama: It's fine...it's a small utility function so it should be relatively safe. We're not holding on to any references or creating circular references, and we're not adding additional *stuff* to existing objects, so this should be fine.

PR #358, opened 2018-06-04
VAM-736: Add a 'ALL VISTA ONLY' configuration option for setting all RPCs to 'VISTA ONLY'

creindle:


emulation/dispatchers/pureEmulation.js
```diff
@@ -245,6 +247,13 @@ class PureEmulationDispatcher {
         });
         return error;
     }
+    static validateDataSource(model, fanout, vistaOnlyRPCs, noEmulateRPCs) {
```
mfuroyama: Just a minor, slightly pedantic comment: consider giving your function here a slightly different name, slightly that more accurately describes the intent of the function. `validateDataSource` implies that there is a "valid-ness" check that you're applying here, which is not the case.

This is not a huge deal per se, but picking good function and variable names goes a long way toward crafting self-documenting code and reducing cognitive overhead for people reading your code.

creindle: Would you say "setDatasource" is a good function name?

PR #367, opened 2018-06-21
VAM-746: Fix the "default context" object in the VICS Server Session service

creindle:
[X] Added `stationNumber` field to all the configurations in `config/nodeEnv.json`
[X] In `emulation/server/config.js`, replaced default object `sessions` field to contain only the stationId from the added `stationNumber` configuration
[X] Fixed minor typo and added field to pass model validity tests
[X] Passed the following tests:
- test-model-validity
- test-services
- test-models
- test-http
- test-benchmark
- jenkins-test-services
- jenkins-test-models
- jenkins-test-http
- jenkins-test-model-validity
- jenkins-test-regression
- jenkins-test-benchmark


emulation/server/config.js
```diff
@@ -1,16 +1,13 @@
...
 
 const modelConfig = config.models || {};
 
 module.exports = {
     name: 'RPC Emulation Server',
     description: 'RPC Emulation Server',
-    sessions: {
-        default: sessionContextDefault,
-    },
+    sessions: { stationId: config.stationNumber },
```
mfuroyama: I _think_ this needs to be:
```Javascript
sessions: {
    default: { stationId: config.stationNumber },
},
```

creindle: Ok, I'll change it to this and run all the tests again.

PR #368, opened 2018-06-22
Vam 744 reorganize and isolate the vics http server component

kekuso:
Run new http-server with `npm run start-http`


mfuroyama:
Hmm, that's weird. The `.gitignore` file should have prevented any commits in the `httpserver/static` directory. Let's make sure that:
* We don't commit/push anything to `httpserver/static` since that stuff is dynamically generated by the VICSClient
* The `gitignore` file is actually correct

PR #369, opened 2018-06-23
Vam 749 create get users route

sidjtd:
Service Function is hard-coded to only return all users that match contexts stationId.

Currently projection also returns all data fields except for mongo "_id".

Endpoint spec file is now pointing to userMetaService as it should, so now it works!


emulation/httpserver/routes/meta/user/listUsers.yml
```diff
@@ -0,0 +1,302 @@
...
+        items:
+          type: object
+          properties:
+            name:
+              type: string
+              example: 'ALEXANDER,ROBERT'
+            id:
+              type: string
+              example: 'urn:vista:999:200:64'
+            type:
```
mfuroyama: The requirement for this endpoint was only to return a list of the users name and ID (URN). Let's add a projection to filter as such and adjust the documentation and the unit tests accordingly.

services/userMetaService/listUsers-spec.js
```diff
@@ -0,0 +1,21 @@
...
+const modelFunction = require('../userMetaService');
+const serviceTestUtils = require('../tests/serviceTestUtils');
+
+describe('listUsers asynchronous specs', () => {
+    let context;
+    beforeAll(() => {
+        context = serviceTestUtils.beforeAllScript(modelFunction).context;
+    });
+
+    it('listUsers is tested for a valid array response', (done) => {
```
mfuroyama: When you think about it, this unit test really isn't doing much. It's only checking that "an array of objects is being returned". Let's get into the habit of writing our Jasmine unit tests such that the specific behavior of your application code is being tested, not just high-level behaviors.

Specifically, what I mean is, yes we know that the service function should return an array of objects...that's currently being tested. But what about the contents of the objects? Expected length? You know what these results should be and your application code has control over that so we should be testing that too. 

Also, while we're in here, let's not forget to test negative cases too. Specifically, let's maybe add a call to the service function where the context object has an invalid or non-standard station number?

mfuroyama:
Also, @kekuso just merged in his change to split the HTTP server out from the VICS Emulation Server application. Let's make sure that this is properly coordinated when we do the final merge, once all the comments above are addressed. Thanks!

services/userMetaService/listUsers-spec.js
sidjtd: Roger that, I think I was being a bit shy evaluating the contents of each object thinking the projection may want to be adjusted at some point.  But yes to negative testing.

God it!

sidjtd: And now tracking on the Adam HTTP Server split as well

httpserver/routes/meta/user/listUsers.yml
```diff
@@ -0,0 +1,49 @@
...
+  /meta/user/:
+    get:
+      summary: Return all users
+      description: Lookup all users with matching station Id
+      produces:
+        - application/json
+      parameters:
+        - in: query
+          name: search
+          description: search term to use as query on problem data
```
mfuroyama: Let's double-check the parameters listed here in the documentation. I'm guessing that these are left overs from a copy-pasta?

services/userMetaService/listUsers.js
```diff
@@ -0,0 +1,28 @@
+'use strict';
+
+const serviceUtils = require('../serviceUtils');
+
+/**
+ * @summary Returns a list of all existing users based off station Id
+ * @description  -
```
mfuroyama: We can probably remove this `@description` tag since it's empty.

services/userMetaService/listUsers.js
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * @memberof UserMetaService
+ * @instance
+ */
+
+function listUsers(context, callback) {
+    const service = this;
+    const stationId = context.stationId;
+
+    service.find('NewPerson', {
```
mfuroyama: Really minor here, but we don't really need to capture the service `this` in the lexical scope using an external variable. We can remove the `service` definition, and this line can just become `this.find(...`

sidjtd: Ok!

httpserver/routes/meta/user/listUsers.yml
sidjtd: Ah crap, yeah those needed to be removed!!! I left them in to see if removing them would caused issues with doc-http.... removing!

sidjtd: Yes!  I wasn't originally sure if doc-http would run without parameters, but it was ran and it ran fine!!!! The deletion of the parameters is now committed...

PR #372, opened 2018-06-29
Vam 748 user authentication

sidjtd:
3 files: Endpoint, Test-spec, and YML. (4th is just clean up of redundant test line in last function)


httpserver/routes/meta/user/authenticateUser.js
```diff
@@ -0,0 +1,18 @@
...
+
+const sessionManager = require('../../../../services/sessionService/manager');
+const _ = require('lodash');
+
+const authenticateUser = (req, res) => {
+    const sessionId = _.get(req, 'connection.remoteAddress', false);
+    const body = _.get(req, 'body', false);
+
+    sessionManager.add(sessionId, body);
+    return res.json({ success: 'Successfully added Session Id to Context' });
```
mfuroyama: Hmm, good first try, but unfortunately, this code isn't going to accomplish what is expected against the issue requirement:

1. First of all, let's refresh our memory on the [Session Context Object field specs(https://github.com/vistadataproject/VICSServer/wiki/Session-Context-Object). Notice that we need the user's URN set to `userId` in the context object. Given what you've documented in the endpoint YAML file and what the code is doing here, not only will the session object not have the right field, but potentially, it'll hold the users password as well!

2. Let's default the sessionId  to something meaningful, like 'localhost'

3. Instead of just blindly doing an `add` with the `sessionManager`, let's do a `get` and then an `update`. The endpoint should be **updating** the session context object, not creating it.

4. Finally, consider changing the success text here to reflect the eventual intent of the endpoint. True, it's a semi-placeholder for now, but a better message might be `'Successfully authenticated user'`

httpserver/routes/meta/user/authenticateUser.yml
```diff
@@ -0,0 +1,40 @@
...
+      summary: Authenticates user
+      description: Associates Client Address to Current Session 
+      parameters:
+        - in: body
+          name: problem data
+          description: Problem Object contains data required to add new problem to selected user
+          schema:
+            type: object
+            properties:
+              connection:
```
mfuroyama: Let's remove the `connection` portion of the schema. This is preset by ExpressJS, and we have no control over it, so there's no need to document it.

httpserver/routes/meta/user/authenticateUser.yml
```diff
@@ -0,0 +1,40 @@
...
+                  properties:
+                    id:
+                      type: string
+                      example: 'urn:vista:999:200:64'
+                    code:
+                      type: string
+                      example: 'not used at this time'
+      responses:
+        200:
+          description: Session ID added
```
mfuroyama: `Session ID added` => `Authentication successful`

httpserver/routes/meta/user/authenticateUser.yml
sidjtd: Ah, I see, I thought we were documenting the main parameters an endpoint accepts, but that makes sense, if its pre-set then thats how it is!

httpserver/routes/meta/user/authenticateUser.js
```diff
@@ -0,0 +1,19 @@
...
+
+const sessionManager = require('../../../../services/sessionService/manager');
+const _ = require('lodash');
+
+const authenticateUser = (req, res) => {
+    const sessionId = _.get(req, 'connection.remoteAddress', 'localhost');
+    const userId = _.get(req, 'body.id', 'user Id invalid');
+
+    const currentSession = sessionManager.get(sessionId);
+    sessionManager.update(currentSession, userId);
```
mfuroyama: This is still not correct. `update` takes an object and you're passing in a string. The line above should be:
```Javascript
sessionManager.update(currentSession, { userId });
```

httpserver/routes/meta/user/tests/authenticateUser-spec.js
```diff
@@ -0,0 +1,42 @@
...
+        authenticateUser.callback(data.req, data.res);
+        expect(authenticateUser.callback).toHaveBeenCalled();
+        expect(authenticateUser.callback).toHaveBeenCalledWith(data.req, data.res);
+    });
+
+    it('tests "authenticateUser" using potential "failure" criteria', () => {
+        authenticateUser.callback(data.req, data.res);
+        expect(authenticateUser.callback).toHaveBeenCalled();
+        expect(authenticateUser.callback).not.toHaveBeenCalledWith(false, null);
+    });
```
mfuroyama: It looks like you're not stubbing the call to the session manager. Because of that, consider double-checking, via the session service, in the specs here that the `userId` field was, in fact, set in the session context object. If you did that, you probably would have caught the broken code above.

PR #373, opened 2018-07-02
VAM 754 - Convert the VICS Server HTTP Server from HTTP to HTTPS

creindle:
[X] Converted HTTP server to HTTPS
[X] Tested HTTPS server with VICSClient and Postman
[X] [Wiki page completed](https://github.com/vistadataproject/VICSServer/wiki/Convert-the-VICS-Server-HTTP-Server-from-HTTP-to-HTTPS)


httpserver/index.js
```diff
@@ -39,7 +41,15 @@ class VICSServerExpress {
...
             engine: {},
+            headers: {
+                accept: 'https/https',
+            },
         });
+        this.certificate = {
+            key: fs.readFileSync('./key.pem'),
+            cert: fs.readFileSync('./cert.pem'),
+            passphrase: '', // To be filled depending on what was set in certificate generator command
+        };
```
mfuroyama: This is really cool! Great work!

If I could make one suggestion though: let's come up with a way to make the fields in the `certificate` object settable via configuration (i.e. `config.js`), while keeping `key.pem`, `cert.pem` and the empty passphrase as defaults.

I'm thinking that there's a couple reasons for this:
1. We might use and maintain multiple versions of the certificate files for different environments, and we want the ability to specify different files for each of these environments.
2. We don't want to hard-code the passphrase anywhere, and in a real operational environment, we'll probably pass the passphrase through either a parameter or an environment variable setting. Using the configuration settings for this makes this pretty easy.

Let me know if this makes sense or not

creindle: Hey Mike, I added the certificate passphrase as a new environment config variable to `VICSServer/config/index.js` like this 
```javascript
setEnvironmentConfigValue('CERT_PASSPHRASE', 'httpscertificate.passphrase');
```
Then, I added the new configuration to `VICSServer/httpserver/config.js` as 
```javascript
httpscertificate: {
     passphrase: config.httpscertificate.passphrase || '',
},
```
which should allow me to pass the passphrase as an argument to the command `npm run start-http`, correct? I ran the command as `npm run start-http CERT_PASSPHRASE='test'` and it isn't passing through to the `httpserver/config.js` file.

httpserver/config.js
```diff
@@ -1,17 +1,23 @@
...
     logging: config.logging || {},
-    http: {
+    https: {
         port: 8888,
     },
+    httpscertificate: {
+        key: fs.readFileSync('./key.pem'),
+        cert: fs.readFileSync('./cert.pem'),
+        passphrase: config.httpscertificate.passphrase || '', // To be passed in by arguments
+    },
```
mfuroyama: Two things here:
1. Let's just specify the key/cert file names here and hold off on doing the file reads until we access the configuration in `index.js`. This delayed file-read approach gives us the flexibility to specify different file names in the config. 

2. What do you think about this: let's move the `httpscertificate` field into the `https` object. So, the exported config would look something like this:
```javascript
{
    ...
    https: {
        port: 8888,
        certificate: {
            key: './key.pem',
            cert: './cert.pem',
            passphrase: ....
        }
    }
    ...
}
```

httpserver/index.js
```diff
@@ -62,11 +66,13 @@ class VICSServerExpress {
...
 }
 
-// Start the HTTP (ExpressJS) server port
-const httpPort = _.get(config, 'http.port', 8888);
-const http = new VICSServerExpress();
-http.listen(httpPort, () => {
-    log.info(`HTTP Server listening on port ${httpPort}`);
+// Start the HTTPS (ExpressJS) server port
+const httpsPort = _.get(config, 'https.port', 8888);
+const httpsCertificate = _.get(config, 'httpscertificate', {});
```
mfuroyama: ...so, if you implement the ideas above, we'll have to the file reads here

PR #376, opened 2018-08-31
Vam 792 create rpc emulation tester scripts

sidjtd:
generateMetaRPCResults.js tests VICS on port 9011 (or native Vista on any other port of your choosing) when ran.  It creates a metaRPCResults.json in tests/comparisons/Logs.

generateMetaRPCResults.js can be ran again with a different port, and  metaRPCResults.json will be able to retain the responses of 1 VICS (currently 9011) and any other Vista system in the same file.

metaReportGenerator.js uses the JSON to compare discrepancies between responses, as well as calculating elapsed time.  It also is currently set up to ignore any RPCs where RPCs with the 'isDifferent boolean' is set to true. This can be disabled by commenting some lines out.


tests/comparisons/metaReportGenerator.js
```diff
@@ -0,0 +1,133 @@
...
+
+function analyze(rpcObj) {
+    rpcObj.forEach((el, index) => {
+        const mean = _.mean([el.vistaElapsed, el.vicsElapsed]).toPrecision(3);
+        resJSON[index].mean = mean;
+        resJSON[index].ratio = getPercentageChange(el.vicsElapsed, el.vistaElapsed);
+    });
+}
+
+analyze(resJSON);
```
mfuroyama: One suggestion: let's separate and create distinct divisions between the _function definitions_ and the _functional execution_ here. We can do that by taking all the _execution_ code and placing it in a block at the end of the module. Even better, let's wrap that _execution_ code in the following block:

```Javascript
if (require.main === module) {
    \\ Execution code goes here
}
```

tests/comparisons/metaReportGenerator.js
```diff
@@ -0,0 +1,133 @@
...
+    let rows = dedupe
+        .map((el, index) =>
+            `| ${index + 1} | ${el.name} | ${(el.vicsElapsed * 1000.0).toFixed(2)} | ${(el.vistaElapsed * 1000.0).toFixed(2)} | ${(el.mean * 1000.0).toFixed(2)} | ${el.ratio}% |`)
+        .map((lines) => {
+            return `
+${lines}`;
+        });
+    rows.unshift(top);
+    rows = rows.join('');
+    return rows;
```
mfuroyama: Consider cleaning up the code here by reducing to one chain, then joining with a newline:
```Javascript
const rows = _.uniqBy(set, 'name')
    .filter(el => el.name)
    .map((el, index) => `| ${index + 1} | ${el.name} | ${(el.vicsElapsed * 1000.0).toFixed(2)} | ${(el.vistaElapsed * 1000.0).toFixed(2)} | ${(el.mean * 1000.0).toFixed(2)} | ${el.ratio}% |`);

rows.unshift(top);
return rows.join('\n');
```

